PR #5, opened 2020-04-20
always apply_transforms when doing tonanosvg()

anthrotype:
Do that after resolve_use and before ungrouping.


rsheeter:
LGTM. Is it possible to have a test that fails before, passes after?

anthrotype:
Added a test for apply_transforms in tonanosvg(). I also found and fixed another bug. In `resolve_use` we were adding an unqualified `g` element without the namespace prefix, but then `ungroup` (which uses xpath with svg namespace to search for group elements) can't find it.

anthrotype:
not sure why but we seem to be getting different approximations of conics to quadratic beziers on different platforms. In the e9ee53c "transform-nano.svg" I get a diff of 1e-3 (only for a one point in the red heart path) between the results I get on my local macOS and the ones on Travis CI (running Linux 64bit): 

```diff
- 88.94,22.155
+ 88.939,22.155 
```

I will probably use a different shape that doesn't include arcs for the sake of this PR, and investigate that separately. Maybe we can add macOS to the CI matrix? Or maybe for now we just care we can run on Linux?

anthrotype:
I worked around it by using cubic beziers instead of arcs, thus skipping the skia `convertConicsToQuads` that was giving slightly different results on mac/linux.

maybe we should reconsider implementing an arcs-to-cubic bezier like fontTools.svgLib?
https://github.com/fonttools/fonttools/blob/master/Lib/fontTools/svgLib/path/arc.py

unlike skia's conic2quads, it doesn't seem to need a tolerance parameter. Also I belive we should get fewer nodes for arcs with cubics vs quads.

PR #13, opened 2020-04-28
Try to make pre/post mul a bit more explicit

rsheeter:


tests/svg_transform_test.py
```diff
@@ -105,3 +105,22 @@ def test_inverse(self):
...
+
+    def test_product_ordering(self):
+        affine1 = Affine2D.identity().rotate(pi / 2)
+        affine2 = Affine2D.identity().rotate(pi / 2, cx=0, cy=1)
+
+        assert Affine2D.product(affine1, affine2) != Affine2D.product(affine2, affine1)
+
+        # Start at 1,0.
+        # Rotate 90° around 0,0 to get 0,1
+        # Rotate a further 90° around 0,1 to get ... still 0,1
```
anthrotype, line 121: i found this example a bit confusing. Of course rotating a point around itself doesn't change the result. It doesn't matter the angle or what other transformation you concatenate before or after. it's always going to do nothing.
Maybe better to use different transforms that show how concatenating two matrices into one is the same as applying one transform first, then applying the other one to the result of the first.

rsheeter: Will add a test. I like having proof nop really does nop.

anthrotype: sgtm

PR #23, opened 2020-05-18
tests plus fix for mangling order of grouped items

rsheeter:


src/picosvg/svg.py
```diff
@@ -280,13 +280,13 @@ def _resolve_clip_path(self, clip_path_url):
...
         # multiple clip paths leave behind their intersection
-        if len(clip_paths) > 1:
-            return svg_pathops.intersection(*clip_paths)
-        elif clip_paths:
-            return clip_paths[0]
-        return None
+        if not clip_paths:
+            raise ValueError("Cannot combine no clip_paths")
+        if len(clip_paths) == 1:
+            clip_paths[0]
```
anthrotype, line 288: what's this statement doing? was it meant to return?

rsheeter: ...yes!

src/picosvg/svg.py
```diff
@@ -354,7 +354,10 @@ def _ungroup(self, scope_el):
         multi_clips = []
         for group in groups:
             # move groups children up
-            for child in group:
+            # reverse because "for each addnext" effectively reverses
+            children = [c for c in group]
```
anthrotype, line 358:
```suggestion
            children = list(group)
```

src/picosvg/svg.py
```diff
@@ -724,7 +729,8 @@ def fromstring(cls, string):
             string = string.replace("xlink:href", _XLINK_TEMP)
 
         # encode because fromstring dislikes xml encoding decl if input is str
-        tree = etree.fromstring(string.encode('utf-8'))
+        parser = etree.XMLParser(remove_blank_text=True)
```
anthrotype, line 732: nice. will this remove the need for the extra stripping in picosvg module?

Issue #32, opened 2020-05-27
non-path related attributes get lost when applying transform or stroke via svg_pathops


anthrotype:
When we pass an SVGShape through either svg_pathops.transform or svg_pathops.stroke, we need to roundtrip from SVGShape to SkPath and back to SVGShape.
However the problem is, since SkPath only holds info about the path proper, and has no knowledge of fill, opacity, id, etc., we are losing all those extra non-path-related attributes when performing either transform or stroke.

When running transform and stroke operations via skia we need to restore the rest of the SVGShape attributes, copying them from the input original SVGShape. For transform operation, we of course should skip the transform attribute that we just applied; similarly for stroke-related attrs when stroking.

I believe for clipping operations (union, intersection) we probably don't need to bother copying these, since we run these on clip paths which are kind of special and limited sort of paths. It's also unclear from which of the input shapes the result of a union or intersection would inherit from.

rsheeter:
stroke looks to me like it already copies attributes around. transform definitely does not.

rsheeter:
...stroke is missing some copying too, adding a test to demonstrate

PR #35, opened 2020-05-27
Fix circular dependency issue

### LINKED ISSUE #33: circular dependency
We [me I think] have managed to setup a dependency cyle that causes importing modules to be order-dependent. This also causes pytest to work for some files or sets of files and fail for others. This slipped by because the complete pytest everything run succeeds.

**sad**

```
python
Python 3.7.7 (default, Mar 10 2020, 15:43:03)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from picosvg import svg_types
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "src/picosvg/svg_types.py", line 19, in <module>
    from picosvg import svg_pathops
  File "src/picosvg/svg_pathops.py", line 19, in <module>
    from picosvg.svg_types import SVGPath, SVGShape
ImportError: cannot import name 'SVGPath' from 'picosvg.svg_types' (src/picosvg/svg_types.py)
```

**glad**

```
python
Python 3.7.7 (default, Mar 10 2020, 15:43:03)
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from picosvg import svg_pathops
>>> from picosvg import svg_types
>>> 
```

Fix the cycle so this type of nonsense doesn't occur.

rsheeter:
Fixes #33

### DIFF
--- a/src/picosvg/svg_types.py
+++ b/src/picosvg/svg_types.py
@@ -358,11 +358,11 @@ class SVGPath(SVGShape, svg_meta.SVGCommandSeq):
         target.walk(arc_to_cubic_callback)
         return target
 
-    @staticmethod
+    @classmethod
     def from_commands(
-        svg_cmds: Generator[svg_meta.SVGCommand, None, None]
+        cls, svg_cmds: Generator[svg_meta.SVGCommand, None, None]
     ) -> "SVGPath":
-        svg_path = SVGPath()
+        svg_path = cls()
         for cmd, args in svg_cmds:
             svg_path._add_cmd(cmd, *args)
         return svg_path
--- a/src/picosvg/svg_types.py
+++ b/src/picosvg/svg_types.py
@@ -20,6 +20,7 @@ from picosvg import svg_pathops
 from picosvg.arc_to_cubic import arc_to_cubic
 from picosvg.svg_path_iter import parse_svg_path
 from picosvg.svg_transform import Affine2D
+from typing import Generator
 
 
 # Subset of https://www.w3.org/TR/SVG11/painting.html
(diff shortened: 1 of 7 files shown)


src/picosvg/svg_types.py
```diff
@@ -79,25 +80,34 @@ def _visible(fill, opacity):
...
 
     def bounding_box(self) -> Rect:
-        x1, y1, x2, y2 = svg_pathops.bounding_box(self)
+        x1, y1, x2, y2 = svg_pathops.bounding_box(self.as_cmd_seq())
         return Rect(x1, y1, x2 - x1, y2 - y1)
 
-    def apply_transform(self, transform: Affine2D):
-        return svg_pathops.transform(self, transform)
+    def apply_transform(self, transform: Affine2D) -> "SVGPath":
+        cmds = svg_pathops.transform(self.as_cmd_seq(), transform)
```
anthrotype, line 87: this is where all the non-path related attributes get lost (https://github.com/googlefonts/picosvg/issues/32).
Before returning a new SVGPath constructed only from `cmds`, we should copy over to the new SVGPath all the rest of the attributes (excluding `transform` in this case).
Maybe we can define an _instance_ method called, I don't know, `replace_commands` (for want of better name) that takes self and the new commands and extra kwargs (to override specific attributes), and returns a new SVGPath with the new commands and replaced attributes.

src/picosvg/svg_types.py
```diff
@@ -348,6 +358,15 @@ def arc_to_cubic_callback(curr_pos, cmd, args, *_):
         target.walk(arc_to_cubic_callback)
         return target
 
+    @staticmethod
+    def from_commands(
+        svg_cmds: Generator[svg_meta.SVGCommand, None, None]
+    ) -> "SVGPath":
+        svg_path = SVGPath()
```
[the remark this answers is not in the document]
rsheeter: Done

src/picosvg/svg_types.py
rsheeter: +1, but will send a separate PR

PR #54, opened 2020-06-18
convert paths from evenodd to nonzero winding

### LINKED ISSUE #52: fill-rule, clip-rule evenodd not supported, used by twemoji
```shell
W0615 21:19:04.836112 4542711232 nanoemoji.py:163] SVG.parse(font-srcs/twemoji/assets/svg//1f1e8-1f1fe.svg) failed: Unable to process group attrib {'fill-rule': 'evenodd', 'clip-rule': 'evenodd'}
W0615 21:19:13.307007 4542711232 nanoemoji.py:163] SVG.parse(font-srcs/twemoji/assets/svg//1f1f1-1f1f8.svg) failed: Unable to process group attrib {'fill-rule': 'evenodd', 'clip-rule': 'evenodd'}
W0615 21:19:14.117563 4542711232 nanoemoji.py:163] SVG.parse(font-srcs/twemoji/assets/svg//1f1f2-1f1f4.svg) failed: Unable to process group attrib {'fill-rule': 'evenodd', 'clip-rule': 'evenodd'}
I0615 21:19:19.862827 4542711232 nanoemoji.py:504] 3242/3245 inputs prepared successfully
```

### LINKED ISSUE #53: add support for fill-rule="evenodd"
svg paths can be drawn using "evenodd" fill-rule. TrueType and CFF2 fonts always use "nonzero" winding fill rule. When we find an svg path with "evenodd", we need to convert to one that looks the same but uses "winding" so we can draw it onto glyf or CFF2. 

Skia-pathops `Path` objects have a `simplify` method that unions all the path's contours respecting the fillType property of the Path -- this can be either FillType.WINDING (default) or FillType.EVEN_ODD --; it then returns a new simplified Path object where the fillType is now set to WINDING.

We need to run that on an svg path every time we find "evenodd" fill-rule, then drop the fill-rule (so it defaults to "nonzero").

anthrotype:
Fixes #52 and #53


src/picosvg/svg.py
```diff
@@ -290,7 +290,7 @@ def _combine_clip_paths(self, clip_paths) -> SVGPath:
             raise ValueError("Cannot combine no clip_paths")
         if len(clip_paths) == 1:
             return clip_paths[0]
-        return intersection([c.as_cmd_seq() for c in clip_paths])
+        return intersection(clip_paths)
```
rsheeter, line 293: this seems like a bug that should have failed a test but evidently didn't; we should add a test

src/picosvg/svg_types.py
```diff
@@ -174,6 +180,16 @@ def end(self):
     def as_path(self) -> "SVGPath":
         return self
 
+    def remove_overlaps(self, inplace=False, is_clip_path=False) -> "SVGPath":
```
rsheeter, line 183: nit: suggest passing fill rule; remove overlaps having to know if it's a clip path seems odd to me

src/picosvg/svg_pathops.py
```diff
@@ -75,9 +75,18 @@ def _end_path(points) -> SVGCommandGen:
...
 
+_SVG_FILL_RULE_TO_SKIA_FILL_TYPE = {
+    "nonzero": pathops.FillType.WINDING,
+    "evenodd": pathops.FillType.EVEN_ODD,
+}
+
 
-def skia_path(svg_cmds: SVGCommandSeq):
-    sk_path = pathops.Path()
+def skia_path(svg_cmds: SVGCommandSeq, fill_rule: str = "nonzero") -> pathops.Path:
```
rsheeter, line 84: maybe don't default fill_rule here? - I think it's reasonable to insist clients specify

src/picosvg/svg_pathops.py
```diff
@@ -111,6 +120,12 @@ def intersection(*svg_cmd_seqs) -> SVGCommandGen:
     return _do_pathop(pathops.PathOp.INTERSECTION, svg_cmd_seqs)
 
 
+def remove_overlaps(svg_cmds: SVGCommandSeq, fill_rule: str) -> SVGCommandGen:
```
rsheeter, line 123: Somewhere, perhaps here, we should elaborate on what we're doing and why. Maybe cite the article you had about winding, etc. Assume a solid SWE who doesn't know about winding is reading basically.

src/picosvg/svg_types.py
```diff
@@ -512,10 +528,22 @@ def as_path(self) -> SVGPath:
...
+    )
 
 
 def intersection(shapes: Iterable[SVGShape]) -> SVGPath:
     return SVGPath.from_commands(
-        svg_pathops.intersection(*[s.as_cmd_seq() for s in shapes])
+        svg_pathops.intersection(
+            *[
+                s.as_path().remove_overlaps(is_clip_path=True).as_cmd_seq()
+                for s in shapes
```
rsheeter, line 546: suggest helper function for the preparation of args for a bool op; it's gotten elaborate enough to be worth

src/picosvg/svg.py
anthrotype: AH! Now I see what's going on here. I had misunderstood the `SVGShape.as_cmd_seq()` method, the function signature says it returns an `svg_meta.SVGCommandSeq`, which is an alias of `Iterable[SVGCommand]`. I forgot that `SVGPath` class has an `__iter__` method that yields exactly SVGCommands; i thought `as_cmd_seq` would return something like a list of tuples (one per command), but it returns an SVGPath! Which is also an SVGShape, and thus `svg_types.intersection` function works both ways (in the original, and in my modified copy in the line above).

I think as_cmd_seq() function should be changed so that its return value is typed as SVGPath, not the generic `svg_meta.SVGCommandSeq`. Return values should be strict and particular, input parameters can be wide.

anthrotype: so basically we _are_ already testing this line, and it does the correct thing whether we pass clip_paths directly, or make that list comprehension with as_cmd_seq() (which also returns SVGPaths, not generic sequences).

src/picosvg/svg_pathops.py
anthrotype: If I don't default, then we'd have to also change `svg_pathops.transform`, `svg_pathops.stroke`, `svg_pathops.bounding_box` and `svg_pathops.union` and `interesection` as well, to pass on the fill_rule down to `skia_path`. But the first three do not really need/use fill_rule (the result won't change if path was interpreted as nonzero or evenodd); for the latter two (which use _do_pathop), they currently take Iterable[SVGCommandSeq] (with `*` args tuple unpacked); if we also need to pass the extra fill_rule, we need to either pass an additional list of fill_rules (one per SVGCommandSeq), or we change the input to be Iterable[Tuple[SVGCommandSeq, str]]. 

What I currently do is instead pass the fill_rule only when it matters (in remove_overlaps, where we evenodd/nonzero does make a difference). For union/intersection where it _would_ make a difference, I prevent that by first removing overlaps (thus fixing winding) of the shapes I will union/intersect (using clip_rule because I assume those will be used in a clipPath context).

src/picosvg/svg_types.py
anthrotype: I will change to pass `fill_rule` as a parameter, but I will keep it optional (default to None), because an SVGPath (self) already knows its `fill_rule` so it was as well just use that.. If only SVG wasn't so complicated and one needs to use `clip_rule` instead of `fill_rule` attribute when performing union/intersection in a clipPath context.

src/picosvg/svg.py
rsheeter: I believe return as SVGCommandSeq to be the correct return type because it makes clear what you can count on being there. Return as SVGPath introduces questions around how other attributes are handled that aren't relevant to an SVGCommandSeq.

src/picosvg/svg_types.py
rsheeter: fill_rule None seems very odd.

src/picosvg/svg_pathops.py
rsheeter: If the result of skia_path depends on the fill_rule then you should require it be passed explicitly. The complexity is fundamental to the operation and shouldn't be hidden. You know the correct value in the cited calling sites so specification there is correct.

src/picosvg/svg.py
```diff
@@ -448,6 +448,9 @@ def _stroke(self, shape):
         # make a new path that is the stroke
         stroke = shape.as_path().update_path(shape.stroke_commands(self.tolerance))
 
+        # skia stroker returns paths with 'nonzero' winding fill rule
```
rsheeter, line 451: explaining why, ideally citing a link, would be nice

src/picosvg/svg_types.py
```diff
@@ -519,11 +536,18 @@ def as_path(self) -> SVGPath:
         return path
 
 
+def _bool_op_clip_paths(shapes: Iterable[SVGShape]) -> Generator[SVGPath, None, None]:
+    for shape in shapes:
+        path = shape.as_cmd_seq()
+        if shape.clip_rule == "evenodd":
+            yield path.remove_overlaps(fill_rule="evenodd", inplace=True)
+        else:
+            yield path
```
rsheeter, line 545: You assume the return type of as_cmd_seq in an unsafe manner; if we were typed this wouldn't compile.

```
for shape in shapes:
  path = shape.as_path()
  if shape.clip_rule == "evenodd":
    path = path.remove_overlaps
  yield path.as_cmd_seq
```

Return type should be SVGCommandSeq not SVGPath.

src/picosvg/svg.py
anthrotype: this is from empirical observations, as stroked paths returned from skia often contain self-intersecting outlines (which would draw holes if interpreted as "evenodd"). We need to override the original shape's fill_rule after stroking, otherwise if the original path's fill rule (before stroking) was "evenodd" and we don't fix it to "nonzero" after the stroking, then as we later perform the remove_overlaps operation on the "evenodd" filled paths (to fix their winding), we will end up with holes, as the self-intersecting stroked outlines will be mistakenly filled as "evenodd"..

src/picosvg/svg_types.py
anthrotype: remove_overlaps() method itself already calls as_cmd_seq() on self before passing control to svg_pathos, calling again as_cmd_seq() on the result of remove_overlaps seemed unnecessary to me..

src/picosvg/svg_pathops.py
anthrotype: ok.. explicit is better than implicit. I was trying to avoid having to change many call sites.

> we need to either pass an additional list of fill_rules (one per SVGCommandSeq), or we change the input to be Iterable[Tuple[SVGCommandSeq, str]]

What do you suggest for `svg_pathops.union` and `svg_pathops.intersection`? How shall we pass down the fill_rule to skia_path?

anthrotype: There is also the logical imbalance of requiring the fill_rule to make a skia_path, on the one hand, and on the other returning an SVGCommandGen which doesn't know/care about the SkPath.fillType, simply yields svg path commands (assuming "nonzero", because that is how we then go on to use those commands).

anthrotype: in other words, we can't roundtrip an SVGPath through a pathops.Path and back to SVGPath without losing the original fill_rule or the Path.fillType as set by skia operations, because we are reducing the Skia Path to an iterable of SVG path commands.

anthrotype: I guess that wherever we currently `return svg_commands(sk_path)` we should instead return a tuple that also includes the SVG fill_rules from the SkPaths' fillType:

```
return svg_commands(sk_path), svg_fill_rules(sk_path)`
```

rsheeter: We can document the result, either by comment or by type. The minimal solution would be to specify in comments. For example intersection would be commented to returns command sequence based on a nonzero fill rule.

Better would be to have SVGCommandSeq have a fill_rule as one cannot reasonably interpret the commands without it. We could do that in this PR or we could finish this PR by adding comments and file an issue to have fill rule on command sequences in a future PR.

src/picosvg/svg.py
rsheeter: That would make a fine comment :)

src/picosvg/svg_pathops.py
anthrotype: after we perform boolean operations with skia-pathops the resulting paths are always FillType.WINDING, because we call them with the default `fix_winding=True`:
https://github.com/fonttools/skia-pathops/blob/6581155e5151624ae9756392db025810933b74d6/src/python/pathops/operations.py#L38

that's what we want indeed, we never want evenodd paths because OT fonts only support nonzero.
I added assertions to confirm that.

The only thing we currently don't support is converting a Skia Path with fillType.EVEN_ODD to SVG with fill_rule="evenodd", because `svg_commands` return value doesn't contain that information. As long the fill type doesn't change from "nonzero" to "evenodd" while inside Skia, this should be fine.

The results of transform, bounding_box or stroke operations is not affected by the original path's fill_rule property. So we can treat incoming SVG paths as "nonzero" and leave the fill rule of the resulting svg_commands unspecified in the case of transform, or assume it will be "nonzero" for stroked outlines.

anthrotype:
i'll merge now, feel free to fix up later as needed

PR #55, opened 2020-06-19
Avoid crashing where a sane alternative exists

rsheeter:
Trying to define ops with degenerate cases in ways that aren't raise an error.

### DIFF
--- a/src/picosvg/svg_types.py
+++ b/src/picosvg/svg_types.py
@@ -72,19 +72,18 @@ class SVGShape:
 
     def might_paint(self) -> bool:
         """False if we're sure this shape will not paint. True if it *might* paint."""
+
         def _visible(fill, opacity):
             return fill != "none" and opacity != 0
             # we're ignoring fill-opacity
 
-        if _visible(self.stroke, self.stroke_opacity):
-            return True
-
-        if not _visible(self.fill, self.opacity):
+        # if all you do is move the pen around you can't draw
+        if all(c[0].upper() == "M" for c in self.as_cmd_seq()):
             return False
 
-        # if all you do is move the pen around you can't draw
-        return any(c[0].upper() != 'M' for c in self.as_cmd_seq())
-
+        return _visible(self.stroke, self.stroke_opacity) or _visible(
+            self.fill, self.opacity
+        )
 
     def bounding_box(self) -> Rect:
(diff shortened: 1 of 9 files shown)


src/picosvg/svg_types.py
```diff
@@ -70,24 +70,33 @@ def _copy_common_fields(
...
         x1, y1, x2, y2 = svg_pathops.bounding_box(self.as_cmd_seq())
         return Rect(x1, y1, x2 - x1, y2 - y1)
 
     def apply_transform(self, transform: Affine2D) -> "SVGPath":
-        cmds = svg_pathops.transform(self.as_cmd_seq(), transform)
         target = self.as_path()
         if target is self:
             target = copy.deepcopy(target)
+        cmds = (('M', (0, 0)),)
+        if not transform.is_degenerate():
```
anthrotype, line 98: why not do a simple if-else here? Most of the time we're throwing that `cmds` object away, as the transform is rarely degenerate.
E.g.
```python
if transform.is_degenerate():
   cmds = (('M', (0, 0)),)
else:
   cmds = svg_pathops.transform(self.as_cmd_seq(), transform)
```

src/picosvg/svg_types.py
```diff
@@ -70,24 +70,33 @@ def _copy_common_fields(
...
+        """False if we're sure this shape will not paint. True if it *might* paint."""
         def _visible(fill, opacity):
             return fill != "none" and opacity != 0
             # we're ignoring fill-opacity
 
-        return _visible(self.fill, self.opacity) or _visible(
-            self.stroke, self.stroke_opacity
-        )
+        if _visible(self.stroke, self.stroke_opacity):
+            return True
```
anthrotype, line 80: a stroke that only has a move also doesn't paint, no?

src/picosvg/svg_types.py
rsheeter: x = default
if something
  x = something else

is shorter and I like, in a language lacking a signal like `if likely(...)`to have the main branch of the if be the likely path

src/picosvg/svg_types.py
rsheeter: what kind of person would make a stroke with only a move :D

rsheeter: Shuffled; now we return false if all cmds are move

Issue #59, opened 2020-06-29
Support replacing CSS style attributes with explicit SVG attributes


anthrotype:
SVG elements can have `style` attribute that contains CSS styling declarations:

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/style

A bunch of noto-emoji SVGs use this way of declaring the fill, stroke, etc. properties of paths.

E.g. https://github.com/googlefonts/noto-emoji/blob/master/svg/emoji_u1f9d1_1f3fe_200d_1f9bc.svg

Picosvg should take attributes like 

`style="stroke:#757575;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:10;"`

and replace them with the equivalent invidual SVG attributes

`stroke="#757575" stroke-linecap="round" stroke-linejoin="round" stroke-miterlimit"10"`

I guess we just need to split at semicolons, then again at colons, to get the list of key:value pairs (unless I'm missing something).

PR #60, opened 2020-07-06
add support for inline element-specific CSS style attributes

anthrotype:
Fixes https://github.com/googlefonts/picosvg/issues/59

SVG elements can have a `style` attribute containing a string formatted as CSS declaration list.

https://www.w3.org/TR/SVG/styling.html#ElementSpecificStyling

In noto-emoji's emoji_u1f469_1f3fd_200d_1f33e.svg and others, this is used in several SVG shapes, in the root svg element and in the linear and radial gradients' stop-colors.

In this PR, we parse these `style` attributes and replace them with the equivalent SVG element attributes upon converting to picosvg.

SVGShape dataclasses gain a new 'style' attribute so we can parse and retain that if needed. The `SVGShape.apply_style_attribute()` method will parse the embedded CSS declarations and convert them to equivalent SVG element attributes, leaving the style attribute itself empty (or only containing remaining style properties for which no corresponding dataclass fields are currently defined).

### DIFF
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -15,10 +15,11 @@
 import copy
 import dataclasses
 from functools import reduce
+import itertools
 from lxml import etree  # pytype: disable=import-error
 import re
 from typing import List, Optional, Tuple
-from picosvg.svg_meta import ntos, svgns, xlinkns
+from picosvg.svg_meta import ntos, svgns, xlinkns, parse_css_declarations
 from picosvg.svg_types import *
 import numbers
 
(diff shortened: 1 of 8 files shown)


src/picosvg/svg.py
```diff
@@ -324,6 +344,7 @@ def _inherit_nondefault_overflow(attrib, child, attr_name):
             "clip-rule": _inherit_copy,
             "fill": _inherit_copy,
             "fill-rule": _inherit_copy,
+            "style": _inherit_copy,
```
rsheeter, line 347: Don't you have to worry about merging styles? Potentially better to apply them and then merge post-apply?

anthrotype: i'm not sure what you mean by "merge" here.. Do you mean if the parent group has style attribute while the child has not (and would thus inherit it) but at the same time the child has one of the attributes that the parent group's style defines with inlined css? I can see how that could end up with potentially clashing definitions in the child.
Note however that, at least for `topicosvg()` we are first applying the inlined style attributes and then ungrouping. So the concern above would only apply if one calls ungroup() before apply_style_attributes.

PR #61, opened 2020-07-09
add SVGPathPen and draw method to SVGPath to support FontTools Pen API

anthrotype:
This is useful in nanoemoji when converting COLR->SVG, as well as in general for better interoperability with existing libs that support the FontTools pen interface.


anthrotype:
argh. I had forgotten picosvg does not depend on fonttools. Maybe we should keep it that way. Let me think

anthrotype:
addressed in https://github.com/googlefonts/nanoemoji/pull/83

PR #68, opened 2020-07-28
add round_floats method; don't auto-round in svg_meta.path_segment

anthrotype:
Fixes https://github.com/googlefonts/picosvg/issues/65 and #70

PR #73, opened 2020-08-04
Handle z position updates better

### LINKED ISSUE #70: relative move after closed path should be relative to previous move pt, not last contour pt
take this svg path:

```xml
<svg xmlns="http://www.w3.org/2000/svg" width="512" height="512" viewBox="0 0 30 30">
  <path d="m0,0 l0,10 l10,0 z m10,10 l0,10 l10,0 z"/>
</svg>
```

It renders like this in Chrome:

image

After we make relative coordinates absolute in picosvg, this path is converted like this, which is incorrect:

```xml
<svg xmlns="http://www.w3.org/2000/svg" width="512" height="512" viewBox="0 0 30 30">
  <path d="M0,0 L10,10 L0,10 Z M20,20 L30,30 L20,30 Z"/>
</svg>
```

image

The new relative move point needs to be computed not from the last point in the previous (closed) contour, but from its original move point.

rsheeter:
Trying to pass the tests for #68 with a different implementation

Fixes #70


src/picosvg/svg_types.py
```diff
@@ -258,6 +258,7 @@ def callback(curr_xy, cmd, args, prev_xy, prev_cmd, prev_args)
           returns sequence of (new_cmd, new_args) that replace cmd, args
         """
         curr_pos = Point()
+        return_pos = curr_pos  # where a z will take you
```
anthrotype, line 261: you don't like `move_pos`? How about `start_pos` then? It's the start of the current sub-path.

src/picosvg/svg_types.py
```diff
@@ -285,15 +286,20 @@ def callback(curr_xy, cmd, args, prev_xy, prev_cmd, prev_args)
...
                     new_y += new_cmd_args[y_coord_idxs[-1]]
 
+                # z takes the pen to the return_pos
+                if new_cmd.lower() == "z":
+                    new_x, new_y = return_pos
+
                 prev_pos = copy.copy(curr_pos)
                 curr_pos = Point(new_x, new_y)
+                if new_cmd.upper() == "M":
+                    return_pos = copy.copy(curr_pos)
```
anthrotype, line 296: I believe picosvg Points are immutable tuples so we don't need the extra copy.

tests/svg_types_test.py
```diff
@@ -33,10 +33,30 @@
         ("m2,2 c1,-1 2,4 3,3 C4 4 5 5 6 6", "M2,2 C3,1 4,6 5,5 C4,4 5,5 6,6"),
         # Elliptic arc that goes haywire when stroked
         ("M7,5 a3,1 0,0,0 0,-3 a3,3 0 0 1 -4,2", "M7,5 A3 1 0 0 0 7,2 A3 3 0 0 1 3,4"),
+        # clock hand's path in which the last point must be == start point when absolutized
+        (
+            "m63.8 63.98h0.4c2.1 0 3.8-1.7 3.8-3.8v-32.4c0-2.1-1.7-3.8-3.8-3.8h-0.4"
+            "c-2.1 0-3.8 1.7-3.8 3.8v32.4c0 2.1 1.7 3.8 3.8 3.8z",
+            "M63.8,63.98 H64.2 C66.3,63.98 68,62.28 68,60.18 V27.78 "
+            "C68,25.68 66.3,23.98 64.2,23.98 H63.8 C61.7,23.98 60,25.68 60,27.78 "
+            "V60.18 C60,62.28 61.7,63.98 63.8,63.98 Z",
```
anthrotype, line 42: this test only passes because we do `round_floats(3)`. If you try to remove the rounding, you'll see the last point before the final Z is `63.800000000000004,63.98 Z` (instead of `63.8,63.98 Z` above). You'll also notice that if you try to run the nanoemoji tests with this updated (non-rounding) picosvg, they will fail for the same reason (ie. you'll see a new duplicate point appearing at the end of some contours).
That's precisely what commit https://github.com/googlefonts/picosvg/pull/68/commits/e83f31583b343f8a26264b22424c59a01a3a197f in my other PR https://github.com/googlefonts/picosvg/pull/68 is meant to address.

src/picosvg/svg_types.py
rsheeter: Fixed

src/picosvg/svg_types.py
rsheeter: +1 for start_pos. In fact I might go all the way to subpath_start_pos!

src/picosvg/svg_types.py
```diff
@@ -285,15 +286,20 @@ def callback(curr_xy, cmd, args, prev_xy, prev_cmd, prev_args)
                 if y_coord_idxs:
                     new_y += new_cmd_args[y_coord_idxs[-1]]
 
+                # z takes the pen to the return_pos
+                if new_cmd.lower() == "z":
+                    new_x, new_y = return_pos
+
                 prev_pos = copy.copy(curr_pos)
```
anthrotype, line 293: same here. we don't need copy

rsheeter: heh I have mutable points on the brain! Fixing.

src/picosvg/svg_types.py
```diff
@@ -286,14 +286,13 @@ def callback(curr_xy, cmd, args, prev_xy, prev_cmd, prev_args)
...
                     new_y += new_cmd_args[y_coord_idxs[-1]]
 
-                # z takes the pen to the return_pos
                 if new_cmd.lower() == "z":
-                    new_x, new_y = return_pos
+                    new_x, new_y = subpath_start_pos
 
-                prev_pos = copy.copy(curr_pos)
+                prev_pos = curr_pos
                 curr_pos = Point(new_x, new_y)
```
anthrotype, line 293: you could even swap in one line if you like `prev_pos, curr_pos = curr_pos, Point(new_x, new_y)`

tests/svg_types_test.py
rsheeter: Good catch, ty. I'll need another test in pico then.

src/picosvg/svg_types.py
rsheeter: I always forget that works, I like it!

rsheeter: I can even do new_x, new_y = curr_pos above! Amazing :D

tests/svg_types_test.py
rsheeter: Amended to snap to subpath start per IM. I'm actually wondering if a generalized version makes sense: upon encountering a new point, if it is *very* similar to a point we've seen before snap it to the prior point.

anthrotype: > I'm actually wondering if a generalized version makes sense

I doubt it would be useful. The only reason we are doing for the last point is to avoid the extra implicit closing line. I see no reasons to do it all the time

src/picosvg/geometric_types.py
```diff
@@ -50,6 +50,11 @@ def __add__(self, other: "Vector") -> "Point":
     def round(self, digits: int) -> "Point":
         return Point(round(self.x, digits), round(self.y, digits))
 
+    def tolerent_equals(
```
anthrotype, line 53: spelling

```suggestion
    def tolerant_equals(
```

I would have named this method `almost_equals`

src/picosvg/geometric_types.py
```diff
@@ -50,6 +50,11 @@ def __add__(self, other: "Vector") -> "Point":
...
+        self, other: "Point", tolerence=_DEFAULT_ALMOST_EQUAL_TOLERENCE
+    ) -> bool:
+        return abs(self.x - other.x) <= tolerence and abs(self.y - other.y) <= tolerence
```
anthrotype, line 56: You could use `math.isclose` with given `rel_tol` parameter, it's supposed to do exactly that
https://docs.python.org/3/library/math.html#math.isclose

src/picosvg/svg_types.py
```diff
@@ -23,6 +23,61 @@
...
+
+def _move_endpoint(curr_pos, cmd, cmd_args, new_endpoint):
+    # we need to be able to alter both axes
+    ((cmd, cmd_args),) = _explicit_lines_callback(curr_pos, cmd, cmd_args)
+
+    x_coord_idxs, y_coord_idxs = svg_meta.cmd_coords(cmd)
+    cmd_args = list(cmd_args)  # we'd like to mutate
+    new_x, new_y = new_endpoint
+    if cmd.islower():
+        new_x = new_x - curr_pos.x
```
anthrotype, line 72: I believe this subtraction operation may introduce futher float numerical imprecision..
That's the reason why in my patch I opted to only do this endpoint adjustment inside the absolute() method -- which, by the way, is the origin of the float imprecision, because of the addition of floats to make relative coords absolute.

src/picosvg/svg_types.py
```diff
@@ -251,6 +306,8 @@ def __iter__(self):
     def walk(self, callback):
         """Walk path and call callback to build potentially new commands.
 
+        Walk may tweak the path even if callback is a nop.
```
anthrotype, line 309: I don't know if that's a good idea, that walk also tweaks the paths outside the provided callbacks.. It's supposed to just "walk" with a callback, with the latter doing the actual modifications if any. In my other patch I opted for providing all the data for a callback to perform the snapping of the last point, but did not do it within the walk itself.

src/picosvg/geometric_types.py
anthrotype: Maybe you decided to not use math.isclose because of its handling of comparison near zero (I see in your test case you are starting the path at M0,0). The math.isclose function has a distinct abs_tol parameter for the absolute tolerance (defaulting to 0.0).
Check out the PEP485 (which introduced math.isclose) on why relative comparison is problematic if either value is zero: https://www.python.org/dev/peps/pep-0485/#behavior-near-zero

tests/svg_types_test.py
rsheeter: Identical points will compress better, but probably that's not enough unless we measure significant wins there.

src/picosvg/geometric_types.py
rsheeter: almost_equals sgtm

src/picosvg/geometric_types.py
rsheeter: `math.isclose(self,x, other.x, abs_tol=tolerence)` doesn't seem like it's gaining me anything. I don't mind changing it if you prefer it but I actually like this way better.

src/picosvg/svg_types.py
rsheeter: In retrospect agreed, will make walk just walk.

src/picosvg/svg_types.py
rsheeter: Fair point. I'm tempted to make snap to end its own thing because only doing it in absolute feels odd to me. I think for now I won't :)

tests/fill-rule-evenodd-nano.svg
```diff
@@ -1,8 +1,7 @@
 <svg xmlns="http://www.w3.org/2000/svg" viewBox="-10 -10 220 120">
   <defs/>
-  <path d="M50,0 21,90 98,35 2,35 79,90 z"/>
+  <path d="M50,0 L21,90 L98,35 L2,35 L79,90 Z"/>
```
anthrotype, line 3: so, not only picosvg paths will be absolute, but repeated implicit commands will also be "exploded". This is because iterating over SVGPath always yields exploded commands, and we are now applying `absolute()` to all shapes.
I tried to avoid exactly that when I added round_floats method to SVGPath, by iterating with parse_svg_path(self.d, exploded=False).
Are we ok with this change?

rsheeter: IMHO it's a good change. picosvg is supposed to be simple, explicit absolute commands is just about as simple as you can get. I'm resolving but please reopen if you disagree.

src/picosvg/geometric_types.py
rsheeter: Absent active protest I'm going to leave it this way. Resolving, please re-ping if you disagree.

tests/fill-rule-evenodd-nano.svg
anthrotype: sgtm

PR #80, opened 2020-08-11
Drop group enable-background; it is deprecated

### LINKED ISSUE #74: Ignore enable-background
Deprecated per https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/enable-background, currently blocks picosvg function on some noto emoji svgs. Strip/ignore it.

rsheeter:
Fixes #74

### DIFF
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -205,6 +205,9 @@ class SVG:
             self.elements[idx] = (el, (shape.as_path(),))
         return self
 
+    def _apply_styles(self, el: etree.Element):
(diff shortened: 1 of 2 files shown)


src/picosvg/svg.py
```diff
@@ -386,6 +390,9 @@ def _ungroup(self, scope_el):
         groups = [e for e in self.xpath(f".//svg:g", scope_el)]
         multi_clips = []
         for group in groups:
+            # sneaky little styles, we hates it precious
```
anthrotype, line 393: LOL

PR #82, opened 2020-08-18
Apply clip-paths after converting strokes to paths, not before

### LINKED ISSUE #77: applying clip-paths before stroked paths yields incorrect results
The same necktie emoji from #76 also exhibits another issue in the order in which we apply transformations in picosvg.

The orange oblique stripes are rendered incorrectly because we resolve clip-paths before we apply the strokes; if we convert stroked paths to filled outlines and then clip them they are rendered correctly. See pictures below after applying the following patch:

```diff
diff --git a/src/picosvg/svg.py b/src/picosvg/svg.py
index 5b28bc4..5420165 100644
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -724,10 +724,10 @@ class SVG:
         self.shapes_to_paths(inplace=True)
         self.resolve_use(inplace=True)
         self.apply_transforms(inplace=True)
-        self.apply_clip_paths(inplace=True)
         self.ungroup(inplace=True)
         # stroke after ungroup to apply group strokes properly
         self.strokes_to_paths(inplace=True)
+        self.apply_clip_paths(inplace=True)
         self.evenodd_to_nonzero_winding(inplace=True)
         self.remove_unpainted_shapes(inplace=True)
         self.absolute(inplace=True)
```

The problem is, if I apply that change, two tests related to ungrouping fail and I am not sure why..

The error I got is `ValueError: Need exactly 1 match for //svg:clipPath[@id="merged-clip-0"], got 0`, under `test_topicosvg` with `ungroup-before.svg` and `ungroup-multiple-children-before.svg` test files.

anthrotype:
Fixes #77


anthrotype:
i'm going to merge this as it's fairly trivial change and would like to include in new picosvg release.

Issue #94, opened 2020-09-22
absolute coordinates add 10~20% to size, consider optimizing path strings


anthrotype:
After I rebuild our test color-fonts using latest nanoemoji and picosvg in https://github.com/googlefonts/color-fonts/pull/11, I noticed that the OT-SVG fonts that embed SVGs converted through picosvg (--color_format=picosvg/picosvgz) are now between 10 to 20 % bigger. I think this is a result of picosvg now always converting coordinates to absolute, for all paths, not just those that need to be simplified for various reasons (e.g. processing with other tools like skia-pathops).

Maybe as an optional optimization picosvg could make path `d` strings more compact by using relative coordinates, as well as reducing whitespace between path commands to the mimimum, and similar other size optimizations.

rsheeter:
@EsoHasInsomnia was asking about using picosvg for shrinking svgs which is somewhat related. A shrink mode is worth at least considering.

PR #103, opened 2020-09-30
Better shape reuse identification

rsheeter:
Improved implementation of shape reuse detection.

Annoyingly not working well for real sample data from Noto https://rsheeter.github.io/android_fonts/emoji.html?q=u:1f970 just yet; normalized shapes only match to 1 decimal place. Good chance I've done something silly and introduced enormous numeric instability.

EDIT: at least some of the paths are actually different despite the shapes being visually the same post-normalization. I think there may be ways to solve this but they are well beyond the intended scope of this PR.


src/picosvg/svg_types.py
```diff
@@ -398,9 +411,17 @@ def absolute_callback(subpath_start, curr_pos, cmd, args, *_):
...
             target = copy.deepcopy(self)
-        target.walk(absolute_callback)
+        target.walk(rewrite_callback)
         return target
 
+    def absolute(self, inplace=False) -> "SVGPath":
+        """Returns equivalent path with only absolute commands."""
+        return self._rewrite_path(_relative_to_absolute, inplace)
+
+    def relative(self, inplace=False) -> "SVGPath":
```
anthrotype, line 421: nice, we could use this for #94

anthrotype (commented):
i haven't finished reviewing this, but from a first glance it's looking promising 👍

anthrotype (commented):
what do you plan to do with the failing tests? Do we want to try re-exporting the SVGs from AI with greater float precision?

src/picosvg/svg_reuse.py
```diff
@@ -70,3 +89,195 @@ def affine_between(
...
+            yield Vector(0.0, 0.0)
+        else:
+            yield Vector(args[x_coord_idxs[-1]], args[y_coord_idxs[-1]])
+
+
+def _nth_vector(path: SVGPath, n: int) -> Vector:
+    vectors = _vectors(path)
+    for _ in range(n):
+        next(vectors)
+    return next(vectors)
```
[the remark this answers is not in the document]
rsheeter: Cool, updated

src/picosvg/svg_reuse.py
```diff
@@ -70,3 +89,195 @@ def affine_between(
...
+
+    # Make path relative, with first coord at 0,0
+    x, y = _first_move(path)
+    path.move(-x, -y, inplace=True)
+
+    # By normalizing vector 1 to [1 0] and making first move off y positive we
+    # normalize away rotation, scale and shear.
+    vec1 = _nth_vector(path, 1)  # ignore M 0,0
+    path.walk(
+        lambda *args: _affine_callback(_affine_vec2vec(vec1, Vector(1, 0)), *args)
```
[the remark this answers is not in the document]
rsheeter: Yes :)

src/picosvg/svg_reuse.py
```diff
@@ -70,3 +89,195 @@ def affine_between(
...
+    Intended use is to normalize multiple shapes to identify opportunity for reuse."""
+
+    path = _affine_friendly(dataclasses.replace(shape, id=""))
+
+    # Make path relative, with first coord at 0,0
+    x, y = _first_move(path)
+    path.move(-x, -y, inplace=True)
+
+    # By normalizing vector 1 to [1 0] and making first move off y positive we
+    # normalize away rotation, scale and shear.
```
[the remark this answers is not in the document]
rsheeter: Good point re: only uniform scale; will add a simple test (probably a square and a rectangle)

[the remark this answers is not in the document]
rsheeter: wdyt of 1) normalize vec1 [1 0] 2) normalize y based on bbox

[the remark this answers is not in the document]
rsheeter: Added manipulation of the first vector with a significant y change

rsheeter:
> what do you plan to do with the failing tests? Do we want to try re-exporting the SVGs from AI with greater float precision?

Upon further investigation some of them are more nefarious than expected: same shape, different drawing commands. I think we'll say Shape Reuse as a project is not done but this PR has moved as far as it intended so I would like to merge.

src/picosvg/svg_types.py
```diff
@@ -507,7 +528,7 @@ def round_floats(self, ndigits: int, inplace=False) -> "SVGPath":
 
         Also reformat the SVGPath.d string floats with the same rounding.
         """
-        target: SVGPath = super().round_floats(ndigits, inplace=inplace)
+        target: SVGPath = super().round_floats(ndigits, inplace=inplace).as_path()
```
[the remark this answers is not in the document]
rsheeter: IIRC it's a pytype issue because super round_floats returns SVGShape; makes sense to me to demand path despite knowing it's nop.

src/picosvg/svg_reuse.py
```diff
@@ -14,40 +14,60 @@
...
+from typing import Generator, Iterable, Optional, Tuple
+from picosvg import svg_meta
 from picosvg.svg_transform import Affine2D
 
 
 # Number of decimal digits to round floats when normalizing or comparing
-DEFAULT_TOLERANCE = 9
+# TODO: hard to get #s to line up well with high tolerance
+# TODO: maybe the input svgs should have higher precision? - only 2 decimals on hearts
+_DEFAULT_TOLERANCE = 6
```
anthrotype, line 31: This corresponds to an _absolute_ tolerance between almost equal numbers. I wonder if we should maybe make this tolerance _relative_ to the size of the viewBox or UPEM? Two given coordinates may be more or less "almost equal" depending on the size of the reference coordinate space.

src/picosvg/svg_types.py
```diff
@@ -239,8 +255,24 @@ def round_floats(self, ndigits: int, inplace=False) -> "SVGShape":
         return target
 
     def almost_equals(self, other: "SVGShape", tolerance: int) -> bool:
```
anthrotype, line 257: what we call `tolerance` here is actually the `ndigits` decimal places for the `round_floats` method. It makes sense to keep `ndigits` for `round_floats`, but less so for an `almost_equals` method. I would keep the current name of the parameter `tolerance` (it's an "absolute" tolerance, really) and as the type instead of an `int` accept a `float`; this would normally be some negative power of 10.

If we do that, then makes sense to also change all the other `tolerance: int` parameters (in `normalize` and `affine_between` public function) to also take a `float` representing the absolute tolerance between two numbers.

src/picosvg/svg_reuse.py
```diff
@@ -70,3 +90,214 @@ def affine_between(
...
+        s = target.norm() / vec.norm()
+
+    affine = Affine2D.compose_ltr((affine, Affine2D.identity().scale(s, s)))
+
+    return affine
+
+
+def _first_y(vectors: Iterable[Vector]) -> Optional[Vector]:
+    for idx, vec in enumerate(vectors):
+        if idx > 0 and abs(vec.y) > 0.1:
```
anthrotype, line 133: again an absolute measure such as 0.1 may be either too big or too small depending of the "magnitude" of the coordinates relative to the viewBox/UPEM

src/picosvg/svg_reuse.py
```diff
@@ -70,3 +90,214 @@ def affine_between(
...
+            new_x = 0
+        if almost_equal(new_y, 0):
+            new_y = 0
+        args[x_coord_idx] = new_x
+        args[y_coord_idx] = new_y
+    return ((cmd, args),)
+
+
+def normalize2(shape: SVGShape, tolerance: int = _DEFAULT_TOLERANCE) -> SVGShape:
+    """Build a version of shape that will compare == to other shapes even if offset.
```
anthrotype, line 175: not just "if offset", but also rotated, scaled, flipped, etc.

src/picosvg/svg_reuse.py
```diff
@@ -70,3 +90,214 @@ def affine_between(
...
+
+    affine = Affine2D.compose_ltr((s1_to_origin, s1_vec1_to_s2_vec2, origin_to_s2))
+    if _try_affine(affine, s1, s2):
+        return _round(affine, s1, s2)
+
+    # Could be non-uniform scaling or mirroring
+    # Try to match up the first y movement
+
+    # Could be non-uniform scaling and/or mirroring
+    # Scale first y movement (after matching up vec1) to match
```
anthrotype, line 266: the two comments above look like they say similar things? one seems redundant

src/picosvg/svg_reuse.py
```diff
@@ -70,3 +90,214 @@ def affine_between(
...
+        return affine
+
+    # Normalize first edge.
+    # Fixes rotation, x-scale, and uniform scaling.
+    s1_vec1 = _nth_vector(s1, 1)
+    s2_vec1 = _nth_vector(s2, 1)
+
+    s1_to_origin = Affine2D.identity().translate(-s1x, -s1y)
+    s2_to_origin = Affine2D.identity().translate(-s2x, -s2y)
+    s1_vec1_to_s2_vec2 = _affine_vec2vec(s1_vec1, s2_vec1)
```
anthrotype, line 253:
```suggestion
    s1_vec1_to_s2_vec1 = _affine_vec2vec(s1_vec1, s2_vec1)
```

src/picosvg/svg_reuse.py
rsheeter: Oh yes that's the old comment! Fixed.

src/picosvg/svg_reuse.py
rsheeter: I don't often comment but when I do I do it twice ... fixed

src/picosvg/svg_reuse.py
```diff
@@ -272,7 +272,7 @@ def _round(affine, s1, s2):
     rotate_s2vec1_off_x = Affine2D.identity().rotate(s2_vec1_angle)
 
     affine = Affine2D.compose_ltr(
-        (s1_to_origin, s1_vec1_to_s2_vec2, rotate_s2vec1_onto_x)
+        (s1_to_origin,  s1_vec1_to_s2_vec1, rotate_s2vec1_onto_x)
```
anthrotype, line 275: remember to `black` it out when you're finished editing

rsheeter: Ran black

src/picosvg/svg_types.py
rsheeter: What's the advantage of passing float 10^-N rather than passing in N?

src/picosvg/svg_reuse.py
rsheeter: The caller can achieve this by providing their own tolerance ... maybe I should take a direct value max_err or some such.

src/picosvg/svg_types.py
anthrotype: i think it conveys better the idea of a "tolerance" as in the "(absolute) difference between two floats". One may want to accept two floats as similar if their absolute difference is within some number which may not necessarily be a multiple of 10**-n

rsheeter: Not use 10^-something?! Who would do that?! Fair point, I think it makes sense for caller to provide a float tolerance then, and perhaps nanoemoji will pick its tolerance based on what it knows about the coordinate system.

src/picosvg/svg_reuse.py
rsheeter: Tolerance now passed in

src/picosvg/svg_reuse.py
rsheeter: Now parameterized

src/picosvg/svg_reuse.py
```diff
@@ -171,7 +116,7 @@ def _affine_callback(affine, subpath_start, curr_pos, cmd, args, *_unused):
     return ((cmd, args),)
 
 
-def normalize2(shape: SVGShape, tolerance: int = _DEFAULT_TOLERANCE) -> SVGShape:
+def normalize(shape: SVGShape, tolerance: float, ndigits: int) -> SVGShape:
```
anthrotype, line 119: do we need both a float `tolerance` and a distinct int `ndigits`? Could we derive the latter from the former somehow?
E.g. we could convert the tolerance float to a string and find the position of the decimal point to get the ndigits:`str(tolerance)[::-1].find(".")`. 
str(float) can return exponential notation for small floats, in that case what follows `1e-...` are the ndigits.

src/picosvg/svg.py
```diff
@@ -765,9 +765,10 @@ def _apply_gradient_translation(self, inplace=False):
                 x_prime = (r1 - c * y_prime) / a
 
                 # sanity check: a`(x`, y`) should be a(x, y)
-                p = affine.map_point((x, y))
+                # all our float brutality damages points; low tolerence sanity checks!
+                p = Point(r1, r2)
                 p_prime = affine_prime.map_point((x_prime, y_prime))
-                assert p.almost_equals(p_prime)
+                assert p.almost_equals(p_prime, tolerence=1e-1), f"{p} != {p_prime}"
```
anthrotype, line 771: typo:

```suggestion
                assert p.almost_equals(p_prime, tolerance=1e-1), f"{p} != {p_prime}"
```

didn't the the relevant test fail?

anthrotype: the misspelling tolerence->tolerance is in picosvg/geometric_types.py, in the `almost_equal` function and the `Point.almost_equal` method, and in the global variable `_DEFAULT_ALMOST_EQUAL_TOLERENCE`

anthrotype: fixed

src/picosvg/svg_reuse.py
rsheeter: Maybe I'll explore snap to nearest tolerance

rsheeter: Done

PR #128, opened 2020-11-09
Preprocess gradientTransform translation

### LINKED ISSUE #127: Apply gradient translation
Noto has things like:

```xml
<linearGradient id="c" x1="63.85" x2="63.85" y1="4245" y2="4137.3" gradientTransform="translate(0 -4122)" gradientUnits="userSpaceOnUse">
<radialGradient id="b" cx="63.523" cy="12368" r="53.477" gradientTransform="matrix(1 0 0 .2631 0 -3150)" gradientUnits="userSpaceOnUse">
```

At minimum, just apply the translate transform on a gradient. Perhaps just apply the whole transform, need to think about that.

Credit to @anthrotype for spotting this pattern.

rsheeter:
Early steps toward surviving some of the weirdness in Noto gradients (weird gradient coords + transforms) better.

Move parsing logic nanoemoji => picosvg to enable.

Fixes #127

### DIFF
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -19,11 +19,19 @@ import itertools
 from lxml import etree  # pytype: disable=import-error
 import re
 from typing import List, Optional, Tuple
-from picosvg.svg_meta import ntos, svgns, xlinkns, parse_css_declarations
+from picosvg.svg_meta import (
+    number_or_percentage,
+    ntos,
+    strip_ns,
+    svgns,
+    xlinkns,
+    parse_css_declarations,
+)
 from picosvg.svg_types import *
+from picosvg.svg_transform import Affine2D
 import numbers
 
-_ELEMENT_CLASSES = {
+_SHAPE_CLASSES = {
     "circle": SVGCircle,
     "ellipse": SVGEllipse,
     "line": SVGLine,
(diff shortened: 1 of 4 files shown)


src/picosvg/svg.py
```diff
@@ -679,17 +702,89 @@ def remove_attributes(self, names, xpath="/svg:svg", inplace=False):
...
+                assert r1 == a * x + c * y + dx
+                assert r2 == b * x + d * y + dy
+                y_prime = (r2 - r1 * b / a) / (d - b * c / a)
+
+                # Sub y` into 1)
+                # 1) x` = (r1 - cy`) / a
+                x_prime = (r1 - c * y_prime) / a
+
+                # sanity check: a`(x`, y`) should be a(x, y)
+                p = affine.map_point((x, y))
```
anthrotype, line 768: nit: `p` is the same as `(r1, r2)` above, no need to recompute here

tests/svg_test.py
```diff
@@ -403,3 +403,47 @@ def test_apply_style_attributes(actual, expected_result):
...
+            '<radialGradient id="b" cx="63.523" cy="12368" r="53.477" gradientTransform="matrix(.5 0 0 .2631 0 -3150)" gradientUnits="userSpaceOnUse"/>',
+            '<radialGradient id="b" cx="63.523" cy="395.366021" r="53.477" gradientTransform="matrix(0.5 0 0 0.2631 0 0)" gradientUnits="userSpaceOnUse"/>',
+        ),
+        # Real example from emoji_u1f44d.svg
+        # Using all 6 parts
+        (
+            '<radialGradient id="d" cx="2459.4" cy="-319.18" r="20.331" gradientTransform="matrix(-1.3883 .0794 -.0374 -.6794 3505.4 -353.39)" gradientUnits="userSpaceOnUse"/>',
+            '<radialGradient id="d" cx="-71.60264" cy="-94.82264" r="20.331" gradientTransform="matrix(-1.3883 0.0794 -0.0374 -0.6794 0 0)" gradientUnits="userSpaceOnUse"/>',
+        ),
+    ],
```
anthrotype, line 434: should you maybe also add a test with gradientUnits="objectBoundingBox" even if relatively less frequent than "userSpaceOnUse"?

src/picosvg/svg.py
```diff
@@ -46,6 +63,10 @@
 _DEFAULT_DEFAULT_TOLERENCE = 0.1
 
 
+# Rounding for rewritten gradient matrices
+_GRADIENT_TRANSFORM_NDIGITS = 6
```
anthrotype, line 67: :+1 I think 6 decimal digits is more than enough

anthrotype (approved):
Thanks for working on this! i think the math checks out.

src/picosvg/svg.py
rsheeter: Ty, fixing

tests/svg_test.py
rsheeter: Done

Issue #46, opened 2020-06-08
Support symbol


rsheeter:
Resolve `<symbol>` when producing a picosvg

anthrotype:
If I look at our noto-emoji svgs, I see that of th 97 files that contain the `<symbol>` element, in 96 of them the symbol has no `id`, so it cannot be xreferenced by a `<use>` element and thus it's as if it doesn't exist; only 1 svg `noto-emoji/svg/emoji_u1f61a.svg` has a symbol with an id that is actually `<use>`d, however it used with a transform that places it way off the parent svg's viewbox so it does not even render...

Now I wonder if we actually need to bother implementing this, also given that `svg-native` format does not support `<symbol>` either, see https://github.com/adobe/svg-native-viewer#supported-features-in-svg-native-in-comparison-to-svg11svg2

> No support for inner <svg> or <symbol> elements.

I'm tempted to simply warn and drop them..

rsheeter:
This may have been an issue with old svg export; I see less symbol fails in the latest noto-emoji (https://github.com/googlefonts/color-fonts/issues/10#issuecomment-725838351). If we're lucky the ones that are left simply need to be re-exported :)

rsheeter:
I guess regardless removing anonymous (or if we feel fancy, unreferenced) symbols seems safe

rsheeter:
I'm now seeing cases where we get Symbols that appear to actually be used in Adobe Illustrator exported svgs so we may need to implement this.

PR #134, opened 2020-11-12
Drop anonymous <symbol>

rsheeter:
Reduces the need to support symbol (#46). Merge after #133 (built on top). Helps with https://github.com/googlefonts/color-fonts/issues/10#issuecomment-725838351.


anthrotype (approved):

tests/drop-anon-symbols-after.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" enable-background="new 0 0 128 128" viewBox="0 0 128 128">
+  <defs/>
```
yisibl, line 2: Why does it always produce empty `defs` element?

PR #137, opened 2020-11-13
Fix bug where strokes don't scale

rsheeter:
As reported in https://github.com/googlefonts/nanoemoji/issues/157


src/picosvg/svg.py
```diff
@@ -371,6 +371,7 @@ def _inherit_nondefault_overflow(attrib, child, attr_name):
             "fill": _inherit_copy,
             "fill-rule": _inherit_copy,
             "style": _inherit_copy,
+            "transform": _inherit_copy,
```
anthrotype, line 374: it's not sufficient to inherit and copy the transform attribute from parent group element to child element like you do here; we need an `_inherit_matrix_multiply` that calls `Affine2D.product(child_transform, group_transform)`, similar to how we do inside `apply_transforms`.

anthrotype: I pushed bc4fbf8 to fix this, PTAL

rsheeter: Oops, that was ... let's politely say silly. Ty for the fix :)

src/picosvg/svg_transform.py
```diff
@@ -65,6 +65,9 @@ def degenerate():
     def fromstring(raw_transform):
         return parse_svg_transform(raw_transform)
 
+    def tostring(self):
+        return f'matrix({" ".join(str(v) for v in self)})'
```
rsheeter, line 69: hm. @anthrotype I wonder if this should be ntos [from svg_meta] not just str

anthrotype: good point, done

PR #145, opened 2020-11-25
setup Github Actions CI

anthrotype:


.github/workflows/test.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install tox
+      run: pip install tox
+    - name: Run the tests
+      run: tox -e py
```
rsheeter, line 38: Why tox? - we don't use tox locally and the way I'm reading the config you are saying run for {python versions} so you have an environment for the right python anyway

anthrotype: because it's better! I tried to explain why in the commit message https://github.com/googlefonts/picosvg/pull/145/commits/8b533eb30aea005b8524830774c7f085f9afda98

rsheeter: Better than what at what? :D

Per IM I suggest documenting what concrete problems it solves in https://googlefonts.github.io/python if this is our new best practice.

rsheeter (approved):

rsheeter:
Can we merge ... I'm dying to have the faster builds for picosvg and nanoemoji lol

anthrotype:
Before we can release we need to add the encrypted pypi token in the GitHub repository settings, under Secrets. I can do that tomorrow

rsheeter:
Actually what might be nice is if you add instructions to https://github.com/python and then I try to follow them to update nanoemoji. Always nice to confirm someone else can do it :)

PR #146, opened 2020-11-26
Add an API to destroy content outside the viewbox

rsheeter:
If entirely outside, goodbye. If partially outside, clipped. Does not attempt to resolve transforms, etc; just destroys content outside. Presumed typical use is to convert to picosvg and only then clip to viewbox.

Meant to power https://github.com/googlefonts/nanoemoji/issues/200.


rsheeter:
The fast CI execution is the best thing ever!

anthrotype (approved):

PR #147, opened 2020-11-26
add automatic release + pypi upload on tagged commits

anthrotype:
We do this in a single workflow so we can make the deploy job dependent on lint+test.
We don't want to accidentally make a broken release if a tag is pushed but the tests do not pass.


anthrotype:
gonna merge and try to push a tag and see what happens

anthrotype:
the good thing is that the `deploy` job waited for both `lint` and `test` jobs to complete first, then it was skipped correctly because this is not a tagged commit.

PR #150, opened 2020-11-27
Use python-version-specific requirements.txt instead of single generic requirements.txt

anthrotype:
We have a bunch of dependencies that are only needed on specific python versions, mostly backports like dataclasses, importlib_resources; others like pytype aren't compatible on latest pythons, etc. 

Right now we use conditional environment markers in either setup.py's `install_requires` or in `dev-requirements.in` to include/exclude named packages for specific py-versions (e.g. `dataclasses; python_version < '3.7'`.
Then we periodically run `pip-compile` (from [pip-tools](https://github.com/jazzband/pip-tools/issues/1205) suite) to "freeze" these abstract, top-level dependencies and produce two generic `requirements.txt` and `dev-requirements.txt`.

The problem is `pip-compile` resolves the environment markers based on the python it is run with, so it produces different results depending on whether it is run on python3.6 or python3.9... That means we often have to go and edit manually the generated `requirements.txt` file generated from pip-compile, which defeats the purpose of it being "generated".

Unfortunately pip-compile does not support generating a combined `requirements.txt` file where differences between python versions/platforms/archs are reconciled through respective environment markers (cf. jazzband/pip-tools#826).

Pip-tools devs currently recommend running pip-compile on each targeted python environment, i.e. generating as many concrete requirements.txt files as are the python environments one wishes to support: jazzband/pip-tools#651

A distinct but related issue is that currently pip-compile's support for setup.py is limited in that [it ignores environment markers](https://github.com/jazzband/pip-tools/issues/1139), whereas it does use them when compiling from a `requirements.in` file.

The following PR attempts to fix thes problems.

1) top-level abstract dependencies (either runtime or develop-time) are moved to external `requirements.in` files; setup.py is changed to read off them so they stay in sync and there's only one place to edit;
2) tox config is modified so that running `tox -e py39` will install deps from `python3.9-requirements.txt`, whereas `tox -e py36` will install from the `python3.6-requirements.txt`, and so forth.
3) add a script `requirements/pip-compile-all.sh` to regenerate all requirements.txt for each supported python, whenever any of the top-level dependencies is added/removed/changed. No more need to tweak them manually.


requirements/pip-compile-all.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+# Run pip-compile to freeze all py3*-requirements.txt files.
+# You should run this every time any top-level requirements in either
+# install-requirements.in or dev-requirements.in are added, removed or changed.
+tox --parallel -e 'py3{6,7,8,9}-requirements'
```
rsheeter, line 5: it's pretty non-obvious that you have to update this file when py3.10 or w/e comes out.

tox.ini
```diff
@@ -6,32 +6,49 @@
 ;     # Only runs the linter checks
 ;   $ tox -e py38
 ;     # Runs all tests against python3.8 only
-;   $ tox -e py
-;     # Runs tests against the python interpreter where tox itself was installed
+;   $ export TOXENV=py39
+;   $ tox
+;     # If present use $TOXENV environment variable
 envlist = lint, py3{6,7,8,9}
```
rsheeter, line 12: We now have at least two enumerations of supported python version, none of which are obvious. Can we avoid that somehow?

rsheeter: Even just a single file that enumerates them that everyone refers to?

rsheeter:
It feels like it approaches being simpler to hand-write the requirements file with the correct version bounds and use CI to detect when you screw up

rsheeter (approved):

tox.ini
anthrotype: the source of truth is now tox.ini's `envlist`. Tox allows to list the environments from its envlist with `tox -l` command. I now iterate over them and run pip-compile for each one.

Issue #142, opened 2020-11-23
add support for fill-opacity


anthrotype:
Currently `fill-opacity` is ignored and dropped by picosvg.
This is evident in the lost right pupil of the FACE WITH MONOCLE

![image](https://user-images.githubusercontent.com/6939968/99947361-ca162200-2d6f-11eb-99ec-cb2778ec5c8c.png)

PR #153, opened 2020-11-30
Add support for fill-opacity

anthrotype:
Fixes https://github.com/googlefonts/picosvg/issues/142

### DIFF
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -777,6 +777,18 @@ class SVG:
 
         return self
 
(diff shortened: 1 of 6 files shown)


anthrotype:
after this path, the FACE WITH MONOCLE regains its eye's pupil

face with monocle

anthrotype:
the mosquito emoji has fill-opacity

image

anthrotype:
mosquito is happy

rsheeter (approved):

PR #161, opened 2020-12-03
remove pinned requirements and always test with latest transitive deps

anthrotype:
The setup was getting overly complicated. It's OK to test the latest which is what the user will get when they pip install us. That way we get notified immediately if anything breaks. And it reduces the information overload for future contributors.


anthrotype:
To know the exact versions that were installed for a given test run one can look at the tox log which lists the exact versions used so one can reproduce if needed:

https://github.com/googlefonts/picosvg/runs/1493966379?check_suite_focus=true#step:6:11

setup.py
```diff
@@ -52,9 +26,17 @@ def readlines(filename):
...
+        "dataclasses>=0.7; python_version < '3.7'",
+        "lxml>=4.0",
+        "skia-pathops>=0.4.1",
+    ],
     extras_require={
-        "dev": develop_deps,
+        "dev": [
+            "pytest",
+            "black==20.8b1",
+            "pytype==2020.11.23; python_version < '3.9'",
```
anthrotype, line 38: I only kept black and pytype pinned to exact versions to avoid surprises with style suddenly formatting differently or typechecker being more or less picky all of a sudden. The dependabot should alert us of new releases for these.

tox.ini
```diff
@@ -17,15 +19,8 @@ skip_missing_interpreters = true
 
 [testenv]
 description = Run pytest against the specified python version
-; if the tox environment name contains `dev` (e.g. `py39-dev`), pip install the
-; current package in editable mode (-e .)
-usedevelop =
-    dev: true
```
anthrotype, line 23: this section is not needed, one can achieve the same thing by passing `tox --develop` option

rsheeter (approved):

anthrotype:
i'll do the same for nanoemoji now

PR #163, opened 2020-12-05
ALlow clip to viewbox when making picosvg

rsheeter:


src/picosvg/picosvg.py
```diff
@@ -50,5 +61,10 @@ def main():
     print(etree.tostring(tree, pretty_print=True).decode("utf-8"))
 
 
+def main():
+    # We don't seem to be __main__ when run as cli tool installed by setuptools
+    app.run(_run)
+
+
 if __name__ == "__main__":
-    main()
```
anthrotype, line 54: this could stay simply `main()`, since the latter will call `app.run(_run)` anyway

anthrotype (approved):
LGTM

PR #169, opened 2020-12-15
Remove filled paths with zero area or stroked paths with zero stroke-width

### LINKED ISSUE #157: remove_unpainted_shapes should drop 0 area unstroked
`<rect x="-1228" y="-57.67" style="fill:#F2F2F2;" width="0" height="0.67"/>` should be dropped by remove_unpainted_shapes; 0 area and no stroke

anthrotype:
Fixes #157

### DIFF
--- a/src/picosvg/svg_pathops.py
+++ b/src/picosvg/svg_pathops.py
@@ -192,9 +192,8 @@ def bounding_box(svg_cmds: SVGCommandSeq):
     return skia_path(svg_cmds, fill_rule="nonzero").bounds
 
 
-def path_area(svg_cmds: SVGCommandSeq, fill_rule: str, simplified=True) -> float:
+def path_area(svg_cmds: SVGCommandSeq, fill_rule: str) -> float:
     """Return the path's absolute area."""
     sk_path = skia_path(svg_cmds, fill_rule=fill_rule)
-    if simplified:
-        sk_path.simplify(fix_winding=True)
+    sk_path.simplify(fix_winding=True)
     return sk_path.area
--- a/src/picosvg/svg_pathops.py
+++ b/src/picosvg/svg_pathops.py
@@ -190,3 +190,11 @@ def stroke(
 
 def bounding_box(svg_cmds: SVGCommandSeq):
     return skia_path(svg_cmds, fill_rule="nonzero").bounds
+
+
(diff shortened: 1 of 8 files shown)


anthrotype:
ping @rsheeter

src/picosvg/svg_pathops.py
```diff
@@ -190,3 +190,11 @@ def stroke(
 
 def bounding_box(svg_cmds: SVGCommandSeq):
     return skia_path(svg_cmds, fill_rule="nonzero").bounds
+
+
+def path_area(svg_cmds: SVGCommandSeq, fill_rule: str, simplified=True) -> float:
```
rsheeter, line 195: You kind of have to dig to find out what simplify means. Suggest comment or just no option, always simplify before computing area.

src/picosvg/svg_types.py
```diff
@@ -199,8 +199,12 @@ def _visible(fill, opacity):
         if all(c[0].upper() == "M" for c in self.as_cmd_seq()):
             return False
 
-        return _visible(shape.stroke, shape.stroke_opacity) or _visible(
-            shape.fill, shape.fill_opacity
+        return (
+            _visible(shape.stroke, shape.stroke_opacity) and shape.stroke_width != 0
+        ) or (
+            _visible(shape.fill, shape.fill_opacity)
+            and svg_pathops.path_area(shape.as_cmd_seq(), fill_rule=shape.fill_rule)
```
rsheeter, line 206: This is becoming rather ugly. I suggest breaking it up, akin to:

```python
# Does it look like the stroke is visible?
if _visible(shape.stroke, shape.stroke_opacity) and shape.stroke_width != 0:
  return True

# No stroke; if the shape is hidden we can't draw
if not _visible(shape.fill, shape.fill_opacity):
   return False

# Only shapes with area paint
return svg_pathops.path_area(shape.as_cmd_seq(), fill_rule=shape.fill_rule) > 0
```

rsheeter (approved):

src/picosvg/svg_pathops.py
anthrotype: I removed the `simplified` option

src/picosvg/svg_types.py
anthrotype: Done

PR #172, opened 2021-02-04
Don't apply blank attributes

rsheeter:
Avoid replacing the default when an attribute is blank to prevent issues like https://github.com/googlefonts/nanoemoji/issues/229.

### DIFF
--- a/tests/svg_test.py
+++ b/tests/svg_test.py
@@ -16,7 +16,7 @@ import dataclasses
 from lxml import etree
 import os
 import pytest
-from picosvg.svg import SVG
+from picosvg.svg import SVG, SVGPath
 from picosvg.svg_meta import strip_ns, parse_css_declarations
 from svg_test_helpers import *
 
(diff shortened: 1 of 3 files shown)


yisibl:
Thanks rsheeter, nice work!

tests/svg_test.py
```diff
@@ -487,3 +487,23 @@ def test_apply_gradient_translation(gradient_string, expected_result):
...
+        # Blank fill
+        # https://github.com/googlefonts/nanoemoji/issues/229
+        (
+            '<path fill="" d=""/>',
+            (SVGPath(),),
+        ),
+    ],
+)
+def test_default_for_blank(svg_content, expected_result):
+    svg_string = (
```
anthrotype, line 504: the svg_test_helpers.py module contains an `svg_string` method that does exactly that (i.e. create an svg root element and interpolate the svg content in it), it's already star-imported at the top, you may want to use it here

anthrotype (approved):

tests/svg_test.py
rsheeter: Good catch, ty

PR #174, opened 2021-02-09
Add editorconfig

yisibl:
Adding the `.editorconfig` file helps to maintain a uniform format and reduce lint errors.

lint-error

Reference template: 
* https://github.com/django/django/blob/master/.editorconfig
* https://github.com/seanfisk/python-project-template/blob/master/.editorconfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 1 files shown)


yisibl:
The `.svg` files should probably use trim_trailing_whitespace=false, which helps us test whitespace.

@rsheeter @anthrotype WDYT?

rsheeter:
Ultimately the rule is "run black" but if this makes it easier to match sgtm

yisibl:
Thanks for the merger, this will achieve better results with "run black". Let black focus on formatting the Python syntax itself.

PR #175, opened 2021-02-09
Fix: Prevent errors when there is whitespace in front of `<?xml ?>`

yisibl:
XMLSyntaxError

```shell
lxml.etree.XMLSyntaxError: XML declaration allowed only at the start of the document
```

### DIFF
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -1016,6 +1016,9 @@ class SVG:
         if isinstance(string, bytes):
             string = string.decode("utf-8")
 
+        # Prevent errors when there is whitespace in front of `<?xml ?>`.
+        string = string.strip()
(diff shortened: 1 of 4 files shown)


yisibl:
@rsheeter PTAL, thanks!

anthrotype:
actaually i don think picosvg should be responsible for dealing with malformed XML. If it doesn't start with the magic xml header, it's just _not_ XML.

anthrotype:
SVG documents don't have to start with the XML header, and they should not contain one if embedded in an HTML page or another SVG. You should be able to strip the whitespace or the whole xml declaration from your SVG files before passing them on to picosvg.

yisibl:
Okay, that can be closed.

PR #178, opened 2021-02-18
Add test doc

yisibl:
Reference: https://googlefonts.github.io/python

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,27 @@ pip install -e .
 picosvg mysvg.svg
 ```
 
+## Test
+
+Install the dev dependencies specified in [`extras_require`](https://github.com/googlefonts/picosvg/blob/main/setup.py#L36-L40).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +25,19 @@ pip install -e .
...
+pip install pytest
+pytest
+```
+
+You can also use [pytest](https://docs.pytest.org/) to test the specified files individually.
+
+```shell
+pytest tests/svg_test.py
+```
+
```
rsheeter, line 40: Best to reference https://googlefonts.github.io/python similar to Releasing. You might need to add this content there. Don't `pip install pytest`; prefer to `pip install -e .[dev]` before running tests

yisibl: Thanks for your review, I updated it.

yisibl: Done, PTAL

PR #193, opened 2021-03-09
fix un-grouping when inherited transform == identity

anthrotype:
When we remove `<g>` elements and matrix-multiply a parent group's transform with a child's own transform, if the result is identity (the two transforms cancel out each other) then we need to ensure we delete the transform attribute on the child element (so it will default to identity). Otherwise things get messed up, e.g. noto-emoji/third_party/region-flags/svg/SV.svg

before:

Screenshot 2021-03-09 at 18 25 41

after:

Screenshot 2021-03-09 at 18 29 08


anthrotype:
change is relatively trivial, so merging now, feel free to comment or fix up later

PR #198, opened 2021-03-19
svg_transform: add decompose_scale and decompose_translation

anthrotype:
Needed in nanoemoji to fix https://github.com/googlefonts/nanoemoji/issues/243

when mapping SVG radial gradients to font space in nanoemoji, we want to pre-apply as much transform as we can to the gradient geometry, thus we need to extract the scale and translation components from an Affine2D.
This PR adds a decompose_scale (copied from SkMatrix) and a decompose_translation (mutuated from Rod's code used for apply gradient translation).

### DIFF
--- a/src/picosvg/geometric_types.py
+++ b/src/picosvg/geometric_types.py
@@ -16,11 +16,11 @@ import math
 from typing import NamedTuple, Optional, Union
 
 
-_DEFAULT_ALMOST_EQUAL_TOLERANCE = 1e-9
+DEFAULT_ALMOST_EQUAL_TOLERANCE = 1e-9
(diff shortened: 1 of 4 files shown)


rsheeter:
We could still end up with over-large numbers in one place or another, no? For sufficiently large #s I would expect we'd have to decompose to a sequence of transforms, or we could just fail such cases.

rsheeter (approved):

anthrotype:
> We could still end up with over-large numbers in one place or another, no?

technically yes, but after this PR it is unlikely, unless the original SVG contained unreasonably large numbers that are out of (visual) bounds even in normal SVG settings. Let's revise if/when do encounter such a problem

Issue #191, opened 2021-03-02
support xlink:href attributes in gradient elements


anthrotype:
the gradients in [emoji_u1f1eb_1f1f0.svg](https://github.com/googlefonts/noto-emoji/blob/svg_flags2/third_party/region-flags/waved-svg/emoji_u1f1eb_1f1f0.svg) use `xlink:href` to inherit from a base gradient (which does set userSpaceOnUse) which contains some stops; they only override some other attributes and have no stops of their own. It's an ingenious approach that we had not encounted yet.

```xml
  <defs>
    <linearGradient id="gradbase" gradientUnits="userSpaceOnUse">
      <stop stop-color="#a43907" stop-opacity="0.9961" offset="0"/>
      <stop stop-color="#fff" offset="1"/>
    </linearGradient>
    <linearGradient id="grad1" xlink:href="#gradbase" x1="826.45044" y1="419.93194" x2="786.12402" y2="396.11032"/>
    <linearGradient id="grad2" xlink:href="#gradbase" x1="851.56865" y1="436.06697" x2="763.03312" y2="419.67318"/>
[...]
```

picosvg should do the right thing and use the inherited attributes and stop elements from the template gradient referenced from xlink:href:

https://www.w3.org/TR/SVG/pservers.html#LinearGradientElementHrefAttribute

yisibl:
I recently reached the end of this problem, Adobe illustrator derived SVG time meeting parallel change and generation `xlink: href`.

anthrotype:
what do you mean? 

Have you tried with this https://github.com/googlefonts/picosvg/pull/199?

yisibl:
I also encountered this problem, #199 can be fixed, thanks.

PR #199, opened 2021-03-25
add method to resolve gradient templates

anthrotype:
Some noto-emoji regional flags (e.g. BZ, FK, GT, MX, NI) contain gradients with `xlink:href` pointing to other gradients used as [templates](https://www.w3.org/TR/SVG/pservers.html#PaintServerTemplates). This PR adds support for resolving these references, copying attributes and stops from the template and removing the template if left orphan.

Fixes https://github.com/googlefonts/picosvg/issues/191


src/picosvg/svg.py
```diff
@@ -944,6 +948,69 @@ def _apply_gradient_translation(self, inplace=False):
...
+        # https://www.w3.org/TR/SVG/pservers.html#PaintServerTemplates
+        if not inplace:
+            svg = SVG(copy.deepcopy(self.svg_root))
+            svg._resolve_gradient_templates(inplace=True)
+            return svg
+
+        href_attr = _xlink_href_attr_name()
+        el_by_id = {el.attrib["id"]: el for el in self.xpath(".//svg:*[@id]")}
+
+        def resolve_gradient_template(gradient):
```
rsheeter, line 963: the nested relatively long function isn't helping with readability much, pop it out?

src/picosvg/svg.py
```diff
@@ -944,6 +948,69 @@ def _apply_gradient_translation(self, inplace=False):
...
+            ref = gradient.attrib[href_attr]
+            if not ref.startswith("#"):
+                raise ValueError(f"Only use #fragment supported, reject {ref}")
+
+            template = el_by_id.get(ref[1:])
```
rsheeter, line 968: suggest ref = ref[1:] once instead of all over

src/picosvg/svg.py
```diff
@@ -944,6 +948,69 @@ def _apply_gradient_translation(self, inplace=False):
...
+                    el = self.resolve_url(shape.fill, "*")
+                except ValueError:  # skip not found
+                    continue
+                if strip_ns(el.tag) not in _GRADIENT_CLASSES:
+                    # unlikely the url target isn't a gradient but I'm not the police
+                    continue
+                used_gradient_ids.add(el.attrib["id"])
+        for grad in self._select_gradients():
+            if grad.attrib.get("id") not in used_gradient_ids:
+                grad.getparent().remove(grad)
```
rsheeter, line 1012: `_safe_remove(grad)`

src/picosvg/svg.py
rsheeter: typing gradient would be nice too

src/picosvg/svg.py
```diff
@@ -944,6 +948,69 @@ def _apply_gradient_translation(self, inplace=False):
...
+
+            # recurse if template references another template
+            if template.attrib.get(href_attr):
+                resolve_gradient_template(template)
+
+            for attr_name in _GRADIENT_ATTRS[strip_ns(gradient.tag)]:
+                if attr_name in template.attrib and attr_name not in gradient.attrib:
+                    gradient.attrib[attr_name] = template.attrib[attr_name]
+
+            if len(gradient) == 0:
```
rsheeter, line 987: len(gradient) threw me; maybe a comment? - `# only copy stops if we don't have our own` or some such

rsheeter (approved):

src/picosvg/svg.py
anthrotype: The nested function is to be able to recurse

PR #202, opened 2021-04-02
add preserveAspectRatio parameter to Affine2D.rect_to_rect method

anthrotype:
Required for supporting https://github.com/googlefonts/picosvg/issues/195

follows algorithm described at https://www.w3.org/TR/SVG/coords.html#ComputingAViewportsTransform

Issue #200, opened 2021-04-01
transform on element referencing clip-path affect how clipPath is applied


anthrotype:
A clipPath coordinates are interpreted using the current coordinate system of the element that references it via clip-path. An element's coordinates can be transformed via `transform`, but picosvg is ignoring this when applying clip paths, leading to incorrect results as show in the example below.

According to the SVG spec, https://www.w3.org/TR/SVG11/masking.html#EstablishingANewClippingPath

> If clipPathUnits="userSpaceOnUse", the contents of the ‘clipPath’ represent values in the current user coordinate system in place at the time when the ‘clipPath’ element is referenced (i.e., the user coordinate system for the element referencing the ‘clipPath’ element via the ‘clip-path’ property). [...]
If attribute ‘clipPathUnits’ is not specified, then the effect is as if a value of 'userSpaceOnUse' were specified.

(We currently don't support `clipPathUnits="objectBoundingBox"`, so ignore the latter for now)

When a shape has a `transform` attribute, that's equivalent to establishing a new user coordinate system, transformed by the given affine matrix. Now, if that same shape also references a `clipPath` element via `clip-path="url(#...)"` attribute (and the referenced clipPath has userSpaceOnUse), the expected rendering should be that 

1) the shape is intersected with the referenced clipPath
2) the transform is applied the clipped result

However, currently picosvg does the opposite of that. It first applies the transformation to the path coordinates, keeping the referenced clipPath unchanged, and then it intersects the transformed shape with the referenced clip-path.

This produces incorrect results like the example below.

Source SVG:

```xml
<svg xmlns="http://www.w3.org/2000/svg" height="1000" width="1000" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 4 4">
  <clipPath id="clip">
    <rect width="3" height="3"/>
  </clipPath>
  <rect width="4" height="4" fill="none" stroke="red" stroke-width="0.1"/>
  <circle cx="3" cy="3" r="2" transform="translate(-1 -1)" clip-path="url(#clip)"/>
</svg>
```

Correct rendering of source SVG:

<img width="200" alt="transform-clip-path-before" src="https://user-images.githubusercontent.com/6939968/113334339-7f0eab80-931b-11eb-92be-d2d01368af16.png">

Incorrect rendering after SVG converted via picosvg:

[...]

PR #203, opened 2021-04-07
Add method to resolve nested SVG elements

### LINKED ISSUE #195: handle nested svg elements
the noto-emoji regional flag of Slovenia ([SV.svg](https://github.com/googlefonts/noto-emoji/blob/svg_flags2/third_party/region-flags/svg/SI.svg)) contains such an `<svg>` element nested inside the root `<svg>`, which establishes a new viewport and user-coordinate system.
Picosvg needs to be taught how to resolve these.

placing these scattered resources below as reminder to self:
https://www.w3.org/TR/SVG/coords.html
https://www.sarasoueidan.com/blog/nesting-svgs/

anthrotype:
Meant to fix #195 

This PR depends and is based off #202 which has to be merged first.

It adds a method to recursively replace all nested SVGs with equivalent `<g>` groups with a `transform` attribute that combines the translation (given by the inner svg's x and y properties) and the scale + translation given by potential new `viewBox` of the inner svg.

It does _not_ do two things that pertain to nested SVGs:
1) resolve percentage units relative to the nearest ancestor SVG viewport
2) apply a clip rectangle the size of the new inner viewport (blocked on #200).

However, this already does fix the issue with the Noto Emoji's Slovenian flag that contains a nested svg element for its emblem (added as test below).


src/picosvg/svg.py
```diff
@@ -891,6 +889,99 @@ def normalize_opacity(self, inplace=False):
...
+
+        # first recurse to un-nest any nested nested SVGs
+        self._swap_elements(
+            (el, self._unnest_svg(el, viewbox.w, viewbox.h))
+            for el in self._iter_nested_svgs(svg)
+        )
+
+        g = etree.Element(f"{{{svgns()}}}g")
+        for el in svg:
+            g.append(el)
```
rsheeter, line 928: How list-like is g? Can you g.extend(svg)?

src/picosvg/svg.py
```diff
@@ -891,6 +889,99 @@ def normalize_opacity(self, inplace=False):
...
+            g.attrib["transform"] = transform.tostring()
+
+        # TODO Define a viewport-sized clipPath once transform+clip-path issue is fixed
+        # https://github.com/googlefonts/picosvg/issues/200
+        return (g,)
+
+    def resolve_nested_svgs(self, inplace=False):
+        """Replace nested <svg> elements with equivalent <g> with a transform.
+
+        NOTE: currently this is still missing two features:
```
rsheeter, line 951: Should clarify what it does if it encounters the unsupported things: not-quite-right answer or error?

src/picosvg/svg_meta.py
```diff
@@ -170,3 +172,10 @@ def parse_css_declarations(
         elif declaration.strip():
             raise ValueError(f"Invalid CSS declaration syntax: {declaration}")
     return "; ".join(unparsed) + ";" if unparsed else ""
+
+
+def parse_view_box(s: str) -> Rect:
+    box = tuple(float(v) for v in re.split(r",|\s+", s))
+    if len(box) != 4:
+        raise ValueError("Unable to parse viewBox")
```
rsheeter, line 180: suggest inclusion of the string you couldn't parse it from

rsheeter (approved):

src/picosvg/svg.py
anthrotype: good point, yes extend does work, i'll change it

src/picosvg/svg.py
anthrotype: no error but not-quite-right-answer (we don't clip and don't resolve percents relative to the correct ancestor)

PR #204, opened 2021-04-07
some flags contain the old HTML <comment> elements, strip that too

anthrotype:
namely the Puertorican flag (PR.svg)

More info on the old comment tag https://html.com/tags/old-comment-tag/

(btw, I'm getting there, just 4 flags to go..)


rsheeter:
Fascinating, never heard of this element before!

anthrotype:
me neither

PR #208, opened 2021-04-15
also inherit root <svg> attributes when un-grouping

anthrotype:
Currently we were only handling attribute inheritance from `<g>` elements to their children. But one can define some inheritable attributes on the root `<svg>` element as well and expect these be propagated down when we perform the un-grouping.
E.g. an svg that defines a global fill=none because all its paths are stroke-only and have no fill.

Fixes https://github.com/googlefonts/nanoemoji/issues/275


tests/global-fill-none-before.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="170" height="89" viewBox="0 0 170 89" fill="none" xmlns="http://www.w3.org/2000/svg">
```
yisibl, line 1: Thank you for your PR.

Please consider adding the test `fill=""` and `fill="transparent"`

Other LGTM.

anthrotype: > test fill="" and fill="transparent"

anthrotype: thanks

> test fill="" and fill="transparent"

but those are not valid SVG fill values. Also, it's not the content of `fill` attribute that i'm testing here, but the fact that it gets propagated down. I think this is sufficient.

Issue #206, opened 2021-04-12
The conversion path produced redundant anchor points


yisibl:
![image](https://user-images.githubusercontent.com/2784308/114353838-1091df00-9ba0-11eb-9f31-d008695983d0.png)

Origin:
```svg
<svg width="193px" height="107px" viewBox="0 0 193 107" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round">
        <polyline stroke="green" stroke-width="20" points="10.0039062 96.9609375 96.0761719 10.8886719 182.148437 96.9609375"></polyline>
    </g>
</svg>
```

picosvg output:

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="193px" height="107px" viewBox="0 0 193 107" version="1.1">
  <defs/>
  <path fill="green" d="M2.933,89.89 L89.005,3.818 Q90.412,2.411 92.249,1.65 Q94.087,0.889 96.076,0.889 Q98.065,0.889 99.903,1.65 Q101.741,2.411 103.147,3.818 L189.22,89.89 Q190.626,91.296 191.387,93.134 Q192.148,94.972 192.148,96.961 Q192.148,98.95 191.387,100.788 Q190.626,102.625 189.219,104.032 Q187.813,105.439 185.975,106.2 Q184.138,106.961 182.148,106.961 Q180.159,106.961 178.322,106.2 Q176.484,105.439 175.077,104.032 L89.005,17.96 L96.076,10.889 L103.147,17.96 L17.075,104.032 Q15.668,105.439 13.831,106.2 Q11.993,106.961 10.004,106.961 Q8.015,106.961 6.177,106.2 Q4.339,105.439 2.933,104.032 Q1.526,102.625 0.765,100.788 Q0.004,98.95 0.004,96.961 Q0.004,94.972 0.765,93.134 Q1.526,91.296 2.933,89.89 Z"/>
</svg>
```

Use Sketch's outline stroke tool
```svg

<svg width="193px" height="107px" viewBox="0 0 193 107" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
[...]

PR #209, opened 2021-04-16
Fix the overlapping problem of inflection point path

yisibl:
image

Fixed: https://github.com/googlefonts/picosvg/issues/206

TODO: Two failed test cases need to be fixed.


tests/svg_test.py
```diff
@@ -278,6 +277,7 @@ def test_transform(actual, expected_result):
         ("inkscape-noise-before.svg", "inkscape-noise-nano.svg"),
         ("flag-use-before.svg", "flag-use-nano.svg"),
         ("ungroup-transform-before.svg", "ungroup-transform-nano.svg"),
+        ("scale-strokes-before.svg", "scale-strokes-nano.svg"),
```
anthrotype, line 280: why moving this line?

src/picosvg/svg_pathops.py
```diff
@@ -183,6 +183,7 @@ def stroke(
 
     # nuke any conics that snuck in (e.g. with stroke-linecap="round")
     sk_path.convertConicsToQuads(tolerance)
+    sk_path.simplify(fix_winding=True)
 
     assert sk_path.fillType == pathops.FillType.WINDING
```
anthrotype, line 188: you can now remove this assert, since calling simplify with fix_winding=True guarantees the fillType is going to be WINDING anyway

tests/scale-strokes-after.svg
```diff
@@ -1,7 +1,7 @@
 <svg xmlns="http://www.w3.org/2000/svg" enable-background="new 0 0 128 128" viewBox="0 0 128 128">
-  <g transform="translate(10, 10) scale(2, 1)">
-    <path d="M8.5,8.5 L8.5,61.5 L61.5,61.5 L61.5,8.5 Z M11.5,11.5 L58.5,11.5 L58.5,58.5 L11.5,58.5 Z"/>
-    <path transform="scale(4,3)" d="M4,4 L4,26 L26,26 L26,4 Z M6,6 L24,6 L24,24 L6,24 Z"/>
-    <path d="M4,4 L4,26 L26,26 L26,4 Z M6,6 L24,6 L24,24 L6,24 Z"/>
-  </g>
+    <g transform="translate(10, 10) scale(2, 1)">
```
anthrotype, line 2: why indentation changed here?

src/picosvg/svg_pathops.py
yisibl: OK

tests/scale-strokes-after.svg
yisibl: We think we still need to add editorconfig to ensure the indentation and newline format of non-Python files.

anthrotype: ok, can you add that in a separate pr please?

tests/svg_test.py
yisibl: Fixed

src/picosvg/svg_pathops.py
yisibl: Done.

tests/scale-strokes-after.svg
yisibl: No problem

src/picosvg/svg_pathops.py
```diff
@@ -183,9 +183,14 @@ def stroke(
...
     # nuke any conics that snuck in (e.g. with stroke-linecap="round")
     sk_path.convertConicsToQuads(tolerance)
-    sk_path.simplify(fix_winding=True)
 
-    assert sk_path.fillType == pathops.FillType.WINDING
+    try:
+        sk_path.simplify(fix_winding=True)
+    except svg_pathops.pathops.PathOpsError:
+        # some tricky paths with very densely packed segments sometimes can trigger a
+        # PathOpsError. We assume they do paint to stay on the safe side.
```
anthrotype, line 191: The comment you copy pasted doesn't make sense in here, replace or remove at least the last sentence

anthrotype (commented):
You need to fix the merge conflict

src/picosvg/svg_pathops.py
yisibl: Okay.

Even if an error is reported here, `sk_path.simplify` is still executed, resulting in inconsistent display of pathops-tricky-path-nano.svg.

anthrotype: ugh, not good. We need to copy the path before we send to simplify and roll back if an exception occur. I'll merge this and fix up, thanks

anthrotype: @yisibl https://github.com/googlefonts/picosvg/commit/727fee175e3523326f15cb02ab4a4e504a8f6e88

PR #221, opened 2021-05-19
add pretty_print parameter to SVG.tostring()

anthrotype:


src/picosvg/picosvg.py
```diff
@@ -52,14 +47,7 @@ def _run(argv):
     if FLAGS.clip_to_viewbox:
         svg.clip_to_viewbox(inplace=True)
 
-    tree = svg.toetree()
-
-    # lxml really likes to retain whitespace
-    for e in tree.iter("*"):
-        e.text = _reduce_text(e.text)
```
anthrotype, line 59: @rsheeter do you remember why we needed this? It seems like XMLParser(remove_blank_text=True) already does what we need, or am I missing something?

rsheeter (approved):

src/picosvg/picosvg.py
rsheeter: I vaguely recall it really liked to keep whitespace around elements or something. If parser with remove blank Just Works that's great.

PR #223, opened 2021-05-20
Restructure processing of clip, transform, and group

### LINKED ISSUE #78: Group opacity with overlapping shapes is not handled well
`picosvg` an svg that uses group opacity and render before/after. They may look quite dissimilar. 

*Sample 1*

Input:
```svg
<svg viewBox="0 0 128 128" xmlns="http://www.w3.org/2000/svg">
  <rect x="24" y="16" width="32" height="112"/>
  <g opacity="0.6">
     <path fill="red" d="M10,30 H100 V120 H10 V30 z"/>
     <path fill="blue" d="M5,25 L105,25 L105,125 L5,125 Z M15,35 L15,115 L95,115 L95,35 Z"/>
  </g>
</svg>
```

Screen Shot 2020-08-10 at 2 02 10 PM

*Sample 2*

Input
```svg
<svg viewBox="0 0 128 128" xmlns="http://www.w3.org/2000/svg">
  <rect x="24" y="16" width="32" height="112" />
  <g opacity="0.7">
     <path fill="red" d="M10,30 H100 V120 H10 V30 z"/>
     <path fill="blue" d="M5,25 L105,25 L105,125 L5,125 Z M15,35 L15,115 L95,115 L95,35 Z"/>
     <g opacity="0.8">
        <path fill="yellow" d="M24,67 c-45,-45 45,-45 0,0 c-45,45 45,45 0,0 z" opacity="0.5"/>
        <path fill="green" d="M10,64 64,32 64,96 Z"/>
     </g>
     <path fill="green" d="M128,64 64,32 64,96  Z"/>
  </g>
</svg>
```

Screen Shot 2020-08-10 at 2 02 31 PM

### LINKED ISSUE #110: we need to set gradientTransform when we apply transforms on paths
I think the way we `apply_transforms` in `topicosvg()` is wrong:
https://github.com/googlefonts/picosvg/blob/09d05fe22ded144bd53af6e58133ce8b146c0588/src/picosvg/svg.py#L565-L598

We are only applying the transforms to the paths, but we are keeping untouched the linear/radial gradients which the transformed paths are filled with. This is not sufficient to produce the same visual result as the original SVG, before picosvg translation.

Take this for example. We have two identical rectangles, that both references the same radialGradient. The rectangles are children of `<g>` nested groups that have some `transform` attribute. One is skewed horizontally by 15 degrees, the other is nested below that and has an additional skewX of -30 degres.

```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
  <defs>
    <radialGradient id="g1" gradientUnits="userSpaceOnUse" cx="64" cy="64" r="24">
      <stop offset="0.1" stop-color="gold" />
      <stop offset="0.95" stop-color="red" />
    </radialGradient>
  </defs>
  <g transform="skewX(15)">
    <rect x="40" y="40" width="48" height="48" fill="url(#g1)"/>
    <g transform="skewX(-30)">
      <rect x="40" y="40" width="48" height="48" fill="url(#g1)"/>
    </g>
  </g>
</svg>
```

The SVG should look like this:

73-69-6d-70-6c-65-5f-72-61-64-69-61-6c

(the PNG above was rasterized using resvg, but Chrome or Firefox produce the same rendering)

After converting the above SVG with picosvg we get the following (incorrect) result.
You can see that the radialGradient is identical to the original SVG and has no `gradientTransform`.
However the paths no longer have the `transform` attribute (it has been applied):

```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
  <defs>
    <radialGradient id="g1" gradientUnits="userSpaceOnUse" cx="64" cy="64" r="24">
      <stop offset="0.1" stop-color="gold"/>
      <stop offset="0.95" stop-color="red"/>
    </radialGradient>
  </defs>
  <path fill="url(#g1)" d="M50.718,40 L98.718,40 L111.58,88 L63.58,88 L50.718,40 Z"/>
  <path fill="url(#g1)" d="M27.624,40 L75.624,40 L60.773,88 L12.773,88 L27.624,40 Z"/>
</svg>
```

image

The correct thing to when applying transforms to shapes that are filled with gradients should be:

1) make a copy of the gradient using a new unique id, since we will modify the gradient and the original gradient may still be referenced by some un-transformed shape;
2) convert  from `gradientUnits="objectBoundingBox"` to "userSpaceOnUse"; this is because after we apply the transforms on the paths, their bounding box will also change, but we want to keep a consistent coordinate space;
3) set the gradient's `gradientTransform` to the final `transform` for the gradient-filled element (i.e. the concatenation of all the parents' transforms)

This is the expected picosvg conversion:

```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
  <defs>
    <radialGradient id="g1" gradientUnits="userSpaceOnUse" cx="64" cy="64" r="24" gradientTransform="matrix(1 0 0.268 1 0 0)">
      <stop offset="0.1" stop-color="gold"/>
      <stop offset="0.95" stop-color="red"/>
    </radialGradient>
    <radialGradient id="g2" gradientUnits="userSpaceOnUse" cx="64" cy="64" r="24" gradientTransform="matrix(1 0 -0.309 1 0 0)">
      <stop offset="0.1" stop-color="gold"/>
      <stop offset="0.95" stop-color="red"/>
    </radialGradient>
  </defs>
  <path fill="url(#g1)" d="M50.718,40 L98.718,40 L111.58,88 L63.58,88 L50.718,40 Z"/>
  <path fill="url(#g2)" d="M27.624,40 L75.624,40 L60.773,88 L12.773,88 L27.624,40 Z"/>
</svg>
```

You can see that it now renders correctly, the same as the original SVG:

s_i_m_p_l_e_5f_r_a_d_i_a_l

***

Another way to "fix" this, now that COLRv1 finally has `PaintTransformed`, would be to _never_ apply transforms for shapes that reference a gradient; but keep the gradient definitions as they are, and wrap the transformed gradient-filled shapes inside a PaintTransformed (which will be applied to both shape and gradient).

### LINKED ISSUE #200: transform on element referencing clip-path affect how clipPath is applied

rsheeter:
Try to correctly handle a somewhat wider range of scenarios. Causes groups to be retained when opacity cannot simply be pushed down. Generate transformed gradients somewhat more correctly.

**Do not submit until nanoemoji is ready to handle groups.**

Fixes #78. Fixes #200. Fixes #110.


README.md
```diff
@@ -8,8 +8,10 @@ Tool to simplify SVGs. Converts an input svg into a "pico" svg:
 
 *   Exactly 1 `<defs>` element, first child of root
 *   Only gradients defined under `<defs>`
-*   Only `<path>` elements without stroke or clipping after the initial `<defs>`
+*   After the initial `<defs>`, only `<g>` and `<path>`
+    * `<g>` is eliminated when possible, but may be retained for opacity
 *   Only absolute coordinates
+*   Only commands that specify full coordinates, no shorthand (H, S, etc)
```
anthrotype, line 14: I don't disagree with that, but why is that related to this change?

anthrotype (commented):
very nice! I only started reviewing this, but I can already tell it's improving things a lot. E.g. the examples in https://github.com/googlefonts/picosvg/issues/110 both work with this patch

src/picosvg/svg.py
```diff
@@ -170,6 +203,41 @@ def _safe_remove(el: etree.Element):
...
+
+def _xpath_for_url(url, el_tag):
+    return f'//svg:{el_tag}[@id="{_id_of_target(url)}"]'
+
+
+def _attrib_to_pass_on(el, current_attrib):
+    attr_catcher = etree.Element("dummy")
+    _inherit_attrib(el.attrib, attr_catcher, skip_unhandled=True)
+    _inherit_attrib(current_attrib, attr_catcher)
+    return {k: v for k, v in attr_catcher.attrib.items()}
```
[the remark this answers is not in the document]
rsheeter: Sick.

README.md
rsheeter: Some of the test diffs are because we used to emit shorthand and now don't

tests/transform-linear-before.svg
```diff
@@ -0,0 +1,26 @@
...
+  </clipPath>
+  <g id="copyme" clip-path="url(#c1)">
+    <ellipse clip-path="url(#c2)"
+             cx="16" cy="16"
+             rx="8" ry="4"
+             fill="url(#lingrad)"/>
+  </g>
+  <g transform="translate(-9,0)">
+    <g transform="rotate(-45 12,16)">
+      <use xlink:href="#copyme" />
```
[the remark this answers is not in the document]
rsheeter: Fixed.

src/picosvg/svg.py
```diff
@@ -389,103 +481,142 @@ def _new_id(self, tag, template):
...
+        return reversed(to_process)
 
-            new_ref_id = new_clip_paths[clip_path.d].attrib["id"]
-            clipped_el.attrib["clip-path"] = f"url(#{new_ref_id})"
+    def _transformed_gradient(self, url, transform):
+        fill_el = self.resolve_url(url, "*")
+        assert _is_gradient(fill_el), f"Not sure how to fill from {fill_el.tag}"
+        new_fill = copy.deepcopy(fill_el)
+        new_fill.attrib["gradientTransform"] = _element_transform(fill_el, transform).round(_GRADIENT_TRANSFORM_NDIGITS).tostring()
+        new_fill_id = new_fill.attrib["id"] + re.sub(r"[^\w\d]", "_", new_fill.attrib["gradientTransform"])
```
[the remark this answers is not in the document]
rsheeter: Good call.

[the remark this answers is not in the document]
rsheeter: Amusingly `_new_id` already does this, should have looked harder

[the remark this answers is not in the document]
rsheeter: Fixed.

rsheeter:
> the examples in #110 both work with this patch

Ty, added a unit test to prove it and marked PR as fixing #110

rsheeter:
For now I have commented out objectBoundingBox tests. I do think we should support them.

src/picosvg/svg.py
```diff
@@ -266,22 +348,21 @@ def apply_style_attributes(self, inplace=False):
...
 
-    def xpath(self, xpath, el=None):
+    def xpath(self, xpath: str, el: etree.Element = None, expected_result_range=None):
         if el is None:
             el = self.svg_root
-        return el.xpath(xpath, namespaces={"svg": svgns()})
+        results = el.xpath(xpath, namespaces={"svg": svgns()})
+        if expected_result_range and len(results) not in expected_result_range:
+            raise ValueError(
+                f"Expected {xpath} matches in {expected_result_range}, {len(results)} results"
```
anthrotype, line 357: I wonder, why is it a range and not just the maximum number of matching elements allowed (1 alwyas being the implicit minimum)?

src/picosvg/svg.py
```diff
@@ -359,14 +440,27 @@ def resolve_use(self, inplace=False):
...
+        self, clip_path_url, transform=Affine2D.identity()
+    ) -> SVGPath:
         clip_path_el = self.resolve_url(clip_path_url, "clipPath")
         self._resolve_use(clip_path_el)
-        self._ungroup(clip_path_el)
 
-        # union all the shapes under the clipPath
-        # Fails if there are any non-shapes under clipPath
-        return SVGPath.from_commands(union([from_element(e) for e in clip_path_el]))
+        transform = _element_transform(clip_path_el, transform)
```
anthrotype, line 449: hm I'm not sure you want to transform the clipPath's `transform` and then its individual children's `transform`s indipendently from one another. I believe the correct thing to do would be, if the clipPath container itself defines a `transform` push that down to the children first (concatenating with the children's transform), and each children's transform with the current transform

src/picosvg/svg.py
```diff
@@ -381,111 +475,155 @@ def _combine_clip_paths(self, clip_paths: Sequence[SVGPath]) -> SVGPath:
             return clip_paths[0]
         return SVGPath.from_commands(intersection(clip_paths))
 
-    def _new_id(self, tag, template):
-        for i in range(100):
+    def _new_id(self, template):
+        for i in range(256):
```
anthrotype, line 479: what if we have more than 256?

src/picosvg/svg.py
rsheeter: Because a range lets you express 0..N cleanly so you can easily say both exactly 1 or 0..1

src/picosvg/svg.py
rsheeter: It was pretty arbitrary ... I can make it substantially larger w/o issue.

rsheeter: 1 << 16 ? ... I just like having an exit rather than looping forever

src/picosvg/svg.py
rsheeter: I'm not quite sure I follow. We pull the clipPath transform above, merge it with any child transform, and then transform the child (below). 

_element_transform is specifically meant for that pattern; you march down the graph and each time you encounter a transform you update the active transform.

Did I muddle that up?

src/picosvg/svg.py
rsheeter: Fixed

src/picosvg/svg.py
anthrotype: actually you're right, I somehow missed the fact that you were re-assigning the local variable `transform`.

src/picosvg/svg.py
anthrotype: even if 65k may seem reasonable now, to me it still feels like an unnecessary restrictions. There's no risk you'll end up in an infinite loop here. It would be better to keep it unbounded, e.g. use `itertools.count` iterator
https://docs.python.org/3/library/itertools.html#itertools.count

but whatever, I don't really mind either way.

src/picosvg/svg.py
```diff
@@ -1103,25 +1148,28 @@ def topicosvg(self, *, ndigits=3, inplace=False):
...
 
+        # Discard useless content
         self.remove_nonsvg_content(inplace=True)
         self.remove_comments(inplace=True)
         self.remove_anonymous_symbols(inplace=True)
         self.remove_title_meta_desc(inplace=True)
+
+        # Simplify things that simplify in isolation
         self.apply_style_attributes(inplace=True)
         self.resolve_nested_svgs(inplace=True)
```
anthrotype, line 1159: once this PR is merged, we can go on and fix that TODO at the end of `_unnest_svg` method. Now that transform and clip-paths work nicely together, we can support the clipping behavior of nested `<svg>` elements: i.e. define a new `clipPath` element containing a rectangle the size of the nested svg viewport and set clip-path attribute on the `<g>` element that we swap in for the nested `<svg>`.

src/picosvg/svg.py
```diff
@@ -1250,18 +1306,26 @@ def _do_not_inherit(*_):
...
+    skip_unhandled: bool = False,
+    skips=frozenset(),
+):
+    attrib = copy.deepcopy(attrib)
     for attr_name in sorted(attrib.keys()):
-        if not attr_name in _GROUP_ATTRIB_HANDLERS:
+        if attr_name in skips:
+            del attrib[attr_name]
+            continue
+        if not attr_name in _INHERIT_ATTRIB_HANDLERS:
```
anthrotype, line 1325: more idiomatic

```suggestion
        if attr_name not in _INHERIT_ATTRIB_HANDLERS:
```

anthrotype (approved):
This is great, in one go it fixes three of the last big outstanding issues in picosvg! 👏 

Left some minor comments below

src/picosvg/svg.py
```diff
@@ -381,111 +479,154 @@ def _combine_clip_paths(self, clip_paths: Sequence[SVGPath]) -> SVGPath:
...
 
-    def _ungroup(self, scope_el):
-        """Push inherited attributes from group down, then remove the group.
+    def breadth_first(self):
+        frontier = [
+            SVGTraverseContext(
+                0,
+                self.svg_root,
+                "/svg[0]",
+                Affine2D.identity(),
```
anthrotype, line 496: technically the root `<svg>` element itself could also have a `transform` attribute, which would mean that the current transform is something else than identity?

src/picosvg/svg.py
```diff
@@ -381,111 +479,154 @@ def _combine_clip_paths(self, clip_paths: Sequence[SVGPath]) -> SVGPath:
...
+            # print(context.path, len(context.clips), "clips")
+            if "clipPath" in context.path:
+                _safe_remove(context.element)
+                continue
+
+            el = context.element
+            _del_attrs(el, "clip-path", "transform")  # handled separately
+
+            skips = {"clip-path", "transform", "opacity"}  # handled separately
+            _inherit_attrib(context.attrib, el, skips=skips)
```
anthrotype, line 564: I see that `transform` attribute is skipped here as it's handled separately; however the function `_attrib_to_pass_on` still collects them, multiply them, and pass them down and stores them into context.attrib; maybe `_attrib_to_pass_on` should use the same list of `skips` so it doesn't do unnecessary work more than once?

src/picosvg/svg.py
anthrotype: I think this is all you need to implement that

```diff
diff --git a/src/picosvg/svg.py b/src/picosvg/svg.py
index 0e45249..749bbca 100644
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -921,9 +921,13 @@ class SVG:
         if transform != Affine2D.identity():
             g.attrib["transform"] = transform.tostring()

-        # TODO Define a viewport-sized clipPath once transform+clip-path issue is fixed
-        # https://github.com/googlefonts/picosvg/issues/200
-        return (g,)
+        clip_path = etree.Element(
+            f"{{{svgns()}}}clipPath", {"id": self._new_id("nested-svg-viewport-%d")}
+        )
+        clip_path.append(to_element(SVGRect(width=width, height=height)))
+        g.attrib["clip-path"] = f"url(#{clip_path.attrib['id']})"
+
+        return (clip_path, g)

     def resolve_nested_svgs(self, inplace=False):
         """Replace nested <svg> elements with equivalent <g> with a transform.
```

anthrotype: FYI, I updated the code above for clipping nested svg; it now supports optional overflow="visible"

src/picosvg/svg.py
rsheeter: Having read that I'm wondering if we shouldn't file a bug specifically about that. However, I don't think we want to try to resolve that here.

src/picosvg/svg.py
```diff
@@ -1058,40 +1098,45 @@ def checkpicosvg(self):
...
-
-            if not any((re.match(pat, el_path) for pat in path_whitelist)):
-                errors.append(f"BadElement: {el_path}")
+        for context in self.breadth_first():
+            if any(context.path.startswith(bp) for bp in bad_paths):
                 continue  # no sense reporting all the children as bad
 
-            el_id = el.attrib.get("id", None)
+            if not any((re.match(pat, context.path) for pat in path_allowlist)):
+                print("Report ", context.path, "known-bad", bad_paths)
```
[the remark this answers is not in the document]
rsheeter: Oh that should have been deleted

src/picosvg/svg.py
```diff
@@ -381,111 +479,154 @@ def _combine_clip_paths(self, clip_paths: Sequence[SVGPath]) -> SVGPath:
...
+                opacity = float(el.attrib.get("opacity", 1.0))
+                assert opacity >= 0 and opacity <= 1
+
+                keep_group = len(el) > 1 and opacity > 0.0 and opacity < 1.0
+
+                if keep_group:
+                    el.attrib.clear()
+                    el.attrib["opacity"] = ntos(opacity)
+                else:
+                    if opacity != 1.0:
```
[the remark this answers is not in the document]
rsheeter: Good call, fixed

src/picosvg/svg.py
```diff
@@ -381,111 +479,154 @@ def _combine_clip_paths(self, clip_paths: Sequence[SVGPath]) -> SVGPath:
...
+                    ]
+
+                if len(paths) != 1 or paths[0] != initial_path:
+                    _replace_el(el, [to_element(p) for p in paths])
+
+            elif _is_group(el.tag):
+                # A group with 0 < opacity < 1 *and* multiple children must be retained
+                # This over-retains groups; no difference unless children overlap
+                opacity = float(el.attrib.get("opacity", 1.0))
+                assert opacity >= 0 and opacity <= 1
```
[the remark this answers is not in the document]
rsheeter: Fixed, ty

src/picosvg/svg.py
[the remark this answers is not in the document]
rsheeter: Sorry, I'm wrong. The number of children is also a factor so it doesn't have to be 0. Changed to < 1.0 in hopes that's clearer.

src/picosvg/svg.py
rsheeter: Included this, fixed the flag svg, discovered a new bug where we would leave behind pointless stroke attributes (stroke-width present but no stroke value), fixed that too. Sick.

src/picosvg/svg.py
rsheeter: #228

src/picosvg/svg.py
rsheeter: Stripped clip-path and transform out as having special handling in traversal.

src/picosvg/svg.py
```diff
@@ -381,111 +479,154 @@ def _combine_clip_paths(self, clip_paths: Sequence[SVGPath]) -> SVGPath:
...
-                clip_paths.append(self._resolve_clip_path(clip_url))
-            el = el.getparent()
+        # Reversed: we want leaves first
+        to_process = reversed(tuple(c for c in self.breadth_first()))
 
-        if not clip_paths:
-            return None
-        return self._combine_clip_paths(clip_paths)
+        for context in to_process:
+            # print(context.path, len(context.clips), "clips")
```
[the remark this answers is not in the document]
rsheeter: yes!

src/picosvg/svg.py
```diff
@@ -381,111 +479,154 @@ def _combine_clip_paths(self, clip_paths: Sequence[SVGPath]) -> SVGPath:
...
+            elif _is_group(el.tag):
+                # A group with 0 < opacity < 1 *and* multiple children must be retained
+                # This over-retains groups; no difference unless children overlap
+                opacity = float(el.attrib.get("opacity", 1.0))
+                assert opacity >= 0 and opacity <= 1
+
+                keep_group = len(el) > 1 and opacity > 0.0 and opacity < 1.0
+
+                if keep_group:
+                    el.attrib.clear()
```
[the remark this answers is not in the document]
rsheeter: I didn't actually clear them off but as opacity being the only acceptable group attribute in picosvg output I think clear is right?

anthrotype:
@googlebot I consent.

anthrotype:
Hm.. looks like dataclasses backport for Python 3.6 does not like when attributes are NamedTuples (like Affine2D) and one uses dataclasses.asdict() on them:

https://github.com/ericvsmith/dataclasses/pull/141
https://github.com/ericvsmith/dataclasses/pull/158

how about we kill python 3.6 support? :)

anthrotype:
(the reason I didn't catch this problem in #230 is because the base branch there was not `main` and CI is configured to only run tests on `main` or on PRs that merge into `main`)

rsheeter:
> how about we kill python 3.6 support? :)

+1

Issue #171, opened 2021-02-01
Error when path is separated by spaces


yisibl:
Original svg

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 750 1125">
    <path d="M448.6,997.7c.2,5.6,16.6,9.4,36.5,8.4s35.8-6.2,35.5-11.8a1.7,1.7,0,0,0-.1-.7c-1.5-5.2-17.3-8.6-36.4-7.7s-34.4,5.8-35.5,11.1Z" fill="green" />
</svg>
```

After [svgo](https://github.com/svg/svgo) compression, the spaces are removed, and an error will be reported:

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 750 1125">
    <path d="M448.6 997.7c.2 5.6 16.6 9.4 36.5 8.4s35.8-6.2 35.5-11.8a1.7 1.7 0 00-.1-.7c-1.5-5.2-17.3-8.6-36.4-7.7s-34.4 5.8-35.5 11.1z" fill="green" />
</svg>
```

```shell
nanoemoji --color_format=glyf_colr_0 $(find ../noto-emoji/svg -name 'emoji_u6a.svg')
Wrote  /Users/yisi/works/nanoemoji/build/config.toml
Generating build.ninja
ninja -C /Users/yisi/works/nanoemoji/build
ninja: Entering directory `/Users/yisi/works/nanoemoji/build'
[2/4] picosvg  --clip_to_viewbox ../../noto-emoji/svg/emoji_u6a.svg > picosvg/regular/emoji_u6a.svg
FAILED: picosvg/regular/emoji_u6a.svg
picosvg  --clip_to_viewbox ../../noto-emoji/svg/emoji_u6a.svg > picosvg/regular/emoji_u6a.svg
Traceback (most recent call last):
  File "/Users/yisi/works/nanoemoji/tutorial-env/bin/picosvg", line 8, in <module>
    sys.exit(main())
  File "/Users/yisi/works/nanoemoji/tutorial-env/lib/python3.8/site-packages/picosvg/picosvg.py", line 66, in main
    app.run(_run, argv=argv)
  File "/Users/yisi/works/nanoemoji/tutorial-env/lib/python3.8/site-packages/absl/app.py", line 303, in run
    _run_main(main, args)
  File "/Users/yisi/works/nanoemoji/tutorial-env/lib/python3.8/site-packages/absl/app.py", line 251, in _run_main
[...]

PR #225, opened 2021-05-21
Fix parsing 'A' large-arc and sweep flags when not separated by space or comma

anthrotype:
Fixes https://github.com/googlefonts/picosvg/issues/171


src/picosvg/svg_path_iter.py
```diff
@@ -13,16 +13,69 @@
...
+    while raw_args:
+        arg = raw_args.pop()
+        # modulo to wrap around
+        converter, regex = arg_types[i % n]
+        m = regex.match(arg)
+        if not m:
+            raise ValueError(f"Invalid argument #{i} for '{cmd}': {arg!r}")
+
+        start, end = m.span()
+        yield converter(arg[start:end])
```
rsheeter, line 71: Why not capture a group?

src/picosvg/svg_path_iter.py
```diff
@@ -13,16 +13,69 @@
...
 _IMPLICIT_REPEAT_CMD = {"m": "l", "M": "L"}
 
 
+def _parse_args(cmd: str, args: str) -> Generator[float, None, None]:
+    raw_args = [s for s in _SEPARATOR_RE.split(args) if s]
+    if not raw_args:
+        return
+
+    # reverse so we can pop()/append() like a stack, quicker than pop(0)/insert(0)
+    raw_args.reverse()
```
rsheeter, line 53: Surely the pop(0) overhead on such tiny lists is immaterial?

anthrotype: maybe, I don't mind removing it, though it seemed uncontroversial to me. popping/inserting from the beginning of a list always reallocates memory

src/picosvg/svg_path_iter.py
anthrotype: I need the span().end to so that I can check if there's any remainder, so I might as well use it?

anthrotype: also I want the whole FLOAT_RE match, not any particular part of it, so a group doesn't seem very useful here

rsheeter (approved):
Feels like we're tiptoing toward writing a proper parser :D

anthrotype:
> Feels like we're tiptoing toward writing a proper parser :D

unless SVG 3.0 makes d strings even crazier, I think with this we cover everything (although I might have to add more tests to convince myself of that). For now, I think this is good enough

src/picosvg/svg_path_iter.py
```diff
@@ -13,16 +13,69 @@
...
+
+    if cmd.upper() == "A":
+        arg_types = _ARC_ARGUMENT_TYPES
+    else:
+        arg_types = ((float, _FLOAT_RE),)
+    n = len(arg_types)
+
+    i = 0
+    while raw_args:
```
rsheeter, line 62: Optional alternative: while i < len(raw_args), instead of pop just read raw_args[i].

anthrotype: I don't think that will work actually. `i` counts the argument number (there are 7 in an arc), raw_args are a list of str (split by comma or space), the length of that list is not necessarily == number of logical argument, as more arguments may be coalesced into a single string (which is why I match the beginning of the arg string, then push the tail back to raw_args so I can parse at the next iteration). I think it's good like this.

anthrotype: ok if I use two distinct iteration variables, `i` and `j` I can do without the pop and push, lemme try

PR #227, opened 2021-05-25
Affine2D: define __matmul__ infix operator; deprecate product()

anthrotype:
The product staticmethod is wrong, instead of matrix-multiplying first x second, it was doing the opposite, second x first.

This PR adds `__matmul__` infix operator ('@') for correct matrix multiplication, and deprecates the old incorrect product() method, temporarily kept only for backward compatibility

### DIFF
--- a/src/picosvg/svg_transform.py
+++ b/src/picosvg/svg_transform.py
@@ -79,7 +79,10 @@ class Affine2D(NamedTuple):
         return f'matrix({" ".join(ntos(v) for v in self)})'
 
     def __matmul__(self, other: "Affine2D") -> "Affine2D":
-        """Returns the product of first × second. Order matters.
+        """Returns the product of self × other. Order matters.
+
+        The combined affine matrix can be thought of mapping by other before applying self.
+
         https://en.wikipedia.org/wiki/Matrix_multiplication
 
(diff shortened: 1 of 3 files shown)


src/picosvg/svg_transform.py
```diff
@@ -77,23 +78,50 @@ def tostring(self):
             return f'translate({", ".join(ntos(v) for v in self.gettranslate())})'
         return f'matrix({" ".join(ntos(v) for v in self)})'
 
-    @staticmethod
-    def product(first: "Affine2D", second: "Affine2D") -> "Affine2D":
-        """Returns the product of first x second.
+    def __matmul__(self, other: "Affine2D") -> "Affine2D":
+        """Returns the product of first × second. Order matters.
```
rsheeter, line 82: I think here you may want to spell out the order to use to achieve "first do this then do that" semantics

anthrotype: but we do already have `compose_ltr` for that, which already has some explanantion why it is useful..

anthrotype: done https://github.com/googlefonts/picosvg/pull/227/commits/928409fbdab3b2ab57d48b1e228ca2ae89eb3d02

PR #230, opened 2021-06-01
Convert "objectBoundingBox" to "userSpaceOnUse" before transforming gradients

### LINKED ISSUE #226: Handle objectBoundingBox gradients
#223 improves processing somewhat but does not result in correct handling of gradients with objectBoundingBox. Uncomment the objectBoundingBox tests in `svg_test.py` and fix. 

Ref https://github.com/googlefonts/nanoemoji/blob/6675c6fd0d11dc7aece5fda779ef25c1b7291071/src/nanoemoji/color_glyph.py#L91-L112

anthrotype:
Fixes #226

(note the base branch is `ungroup`, not `main`)

### DIFF
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -39,6 +39,7 @@ from picosvg.svg_meta import (
     xlinkns,
     parse_css_declarations,
     parse_view_box,
+    _LinkedDefault,
 )
 from picosvg.svg_types import *
 from picosvg.svg_transform import Affine2D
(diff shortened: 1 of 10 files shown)


src/picosvg/svg.py
```diff
@@ -39,6 +39,7 @@
     xlinkns,
     parse_css_declarations,
     parse_view_box,
+    _LinkedDefault,
```
rsheeter, line 42: might as well make it linked_default?

anthrotype:
here's how the test files `transform-{linear,radial}-objectbbox-nano.svg` render after this patch

image


image

src/picosvg/svg.py
anthrotype: it's a class, usually in python you use TitleCase for those

rsheeter: Oh it's a class, then just remove the _?

rsheeter (approved):

src/picosvg/svg.py
anthrotype: ok.. it's just that it's kind of a hack so i'd rather mark it as private (`_` prefixed)

rsheeter: Your very first use is importing it somewhere else so ... it's not *really* private

PR #237, opened 2021-07-17
Fix bug with clippath attrs

### LINKED ISSUE #236: picosvg fails on some noto examples
Failing to picosvg all of Noto blocks testing size implications nanoemoji refactoring (https://github.com/googlefonts/nanoemoji/pull/296) and new Paint formats.

```
picosvg ../color-fonts/font-srcs/noto-emoji/svg/emoji_u1f30d.svg

Traceback (most recent call last):
  File "nanoemoji/venv/bin/picosvg", line 11, in <module>
    load_entry_point('picosvg', 'console_scripts', 'picosvg')()
  File "picosvg/src/picosvg/picosvg.py", line 61, in main
    app.run(_run, argv=argv)
  File "nanoemoji/venv/lib/python3.9/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "nanoemoji/venv/lib/python3.9/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "picosvg/src/picosvg/picosvg.py", line 43, in _run
    svg = SVG.parse(input_file).topicosvg()
  File "picosvg/src/picosvg/svg.py", line 1193, in topicosvg
    svg.topicosvg(inplace=True)
  File "picosvg/src/picosvg/svg.py", line 1212, in topicosvg
    self.simplify(inplace=True)
  File "picosvg/src/picosvg/svg.py", line 666, in simplify
    self._simplify()
  File "picosvg/src/picosvg/svg.py", line 585, in _simplify
    to_process = reversed(tuple(c for c in self.breadth_first()))
  File "picosvg/src/picosvg/svg.py", line 585, in <genexpr>
    to_process = reversed(tuple(c for c in self.breadth_first()))
  File "picosvg/src/picosvg/svg.py", line 541, in breadth_first
    self._resolve_clip_path(child.attrib["clip-path"], transform),
  File "picosvg/src/picosvg/svg.py", line 482, in _resolve_clip_path
    clip_paths = [
  File "picosvg/src/picosvg/svg.py", line 483, in <listcomp>
    from_element(e).apply_transform(_element_transform(e, transform))
  File "picosvg/src/picosvg/svg.py", line 179, in from_element
    raise ValueError(f"Bad tag <{el.tag}>")
ValueError: Bad tag <{http://www.w3.org/2000/svg}g>
```

rsheeter:
Fix bug where a group being erroneously left behind during `<use>` resolution makes clip processing most unhappy. Fixes #236. 

Unnecessary but helpful change: made svg types emit svg w/default valued attributes skipped from string conversion so you see <path d="duck"> instead of SVGPath(...long list of attributes...)


src/picosvg/svg.py
```diff
@@ -1358,13 +1419,26 @@ def _do_not_inherit(*_):
...
 
 
+def _drop_default_attrib(attrib: MutableMapping[str, Any]):
+    for attr_name in sorted(attrib.keys()):
+        value = attrib[attr_name]
+        default_value = attrib_default(attr_name, default=None)
+        if default_value is None:
+            continue
+        if isinstance(default_value, float):
+            value = float(value)
```
anthrotype, line 1429: i don't understand why you cast value to a float when isinstance check already says it is a float?

src/picosvg/svg_meta.py
```diff
@@ -192,3 +192,34 @@ def __new__(cls, attr_name):
 
     def __call__(self, data_obj):
         return getattr(data_obj, self.attr_name)
+
+
+_ATTRIB_DEFAULTS = {
```
anthrotype, line 197: I believe we already listing the attributes' default values in the svg_types.py dataclasses; ideally we'd find a way to remove duplication and reuse

src/picosvg/svg.py
```diff
@@ -1021,106 +1116,74 @@ def resolve_nested_svgs(self, inplace=False):
...
-            svg._resolve_gradient_templates(inplace=True)
-            return svg
+
+        assert _is_gradient(gradient)
 
         href_attr = _xlink_href_attr_name()
+        if href_attr not in gradient.attrib:
+            return  # nop
+
         el_by_id = {el.attrib["id"]: el for el in self.xpath(".//svg:*[@id]")}
```
anthrotype, line 1150: here we are creating this `el_by_id` dictionary (xpath-ing the whole document for element ids) for each gradient, every time we call the new `_apply_gradient_template` method; before this was done only once for all gradients, in `_resolve_gradient_templates`, and use it to quickly look up elements by ids.
I suggest that you either optionally have an `el_by_id` parameter that caller of `_apply_gradient_template` can pass down, or you remove it altogether and do xpath for each gradient to search the element by the ref id

anthrotype (commented):
I find the commit message "Add test to expose bug. Fix bug" is a bit too short/vague; by only reading that, it's unclear what the bug was, besides the PR contains more changes than the bug-fixing

src/picosvg/svg.py
```diff
@@ -76,6 +78,7 @@
 
 
 _ATTRIB_W_CUSTOM_INHERITANCE = frozenset({"clip-path", "transform"})
+_OPACITY_ATTRIB = frozenset({"opacity"})
```
anthrotype, line 81: why is this a set and not a (single) string?

src/picosvg/svg_meta.py
anthrotype: oh sorry I see now you already are reusing the `attrib_default` in svg_types dataclasses, never mind my earlier comment

src/picosvg/svg.py
```diff
@@ -572,27 +638,33 @@ def _transformed_gradient(self, url, transform, shape_bbox):
...
                 continue
 
             el = context.element
-            _del_attrs(el, "clip-path", "transform")  # handled separately
+            _del_attrs(el, *_ATTRIB_W_CUSTOM_INHERITANCE)  # handled separately
 
-            skips = _ATTRIB_W_CUSTOM_INHERITANCE | {"opacity"}  # handled separately
+            skips = _ATTRIB_W_CUSTOM_INHERITANCE | _OPACITY_ATTRIB  # handled separately
+            # context.attrib has inherited attrib
+            # if we inherit atop the existing values we'll double-combine, e.g. opacity will *= opacity
```
anthrotype, line 666: was there a separate bug in addition to #237 where we'd double-combine opacity? if yes, it should figure more prominently in the PR/commit message (and have its own test)

src/picosvg/svg.py
```diff
@@ -572,27 +638,33 @@ def _transformed_gradient(self, url, transform, shape_bbox):
...
+            _del_attrs(el, *(set(context.attrib) - skips))
```
anthrotype, line 667: i'm confused, why we need to call `_del_attrs` here? if you delete context.attribs from the current element before doing `_inherit_attrib`, all of them will be inherited..

rsheeter: The comment meant to explain, essentially the context attributes are the desired attributes, you already mixed in the current ones. Probably I coded that in an unnecessarily obtuse manner :D

src/picosvg/svg.py
rsheeter: Because at one point I thought I'd want to custom-handle fill-opacity and stroke-opacity. It could just be a const str now.

src/picosvg/svg.py
rsheeter: I like option 2, I'll change to that I think.

src/picosvg/svg_meta.py
rsheeter: I moved them.

src/picosvg/svg.py
```diff
@@ -201,10 +200,12 @@ def _is_removable_group(el):
...
     if len(el.attrib) == 0:
         return True
-    return len(el) <= 1 or _opacity(el) in {0.0, 1.0}
+    num_children = sum(1 for e in el if e.tag is not etree.Comment)
 
+    return num_children <= 1 or _opacity(el) in {0.0, 1.0}
 
-def _try_remove_group(group_el):
+
+def _try_remove_group(group_el, push_attrib=False):
```
anthrotype, line 208: looks like `push_attrib` is never used with False? Also, it's not clear from the name what the parameter does, it's actually only "pushing the opacity" from group down to its children, not a generic "attrib". I suggest removing it

anthrotype (approved):
overall LGTM, aside from minor comments above

PR #245, opened 2021-10-07
Fix problem where reuse was normalizing to 0's

rsheeter:
Normalizing to a set of 0's isn't all that helpful. This could occur when the initial drawing command did not move away from the first move, e.g. `M 1,1 L 1,1`.

### DIFF
--- a/reuse_example.html
+++ b/reuse_example.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+<head>
+<style>
+body { margin: 0; }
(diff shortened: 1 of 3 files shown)


anthrotype:
was the `reuse_example.html` added on purpose or committed by accident?

anthrotype:
> See reuse_example.html in root of picosvg for a visual explanation

oh I see it was intentional, sorry for the noise (i'll keep reviewing)

Issue #246, opened 2021-10-12
AssertionError in svg_reuse.py when building noto-emoji COLRv1


anthrotype:
I just [yanked](https://pypi.org/help/#yanked) picosvg==0.18.4 released today, after I encountered an error while building  noto-emoji COLRv1 with latest picosvg + nanoemoji (v0.9.8).

```
  File "/usr/local/google/home/clupo/Github/color-fonts/.venv/lib/python3.9/site-packages/picosvg/svg_reuse.py", line 219, in affine_between
    assert s2_vec1x_idx != -1
AssertionError
```

The two shapes that trigger the assertion error in affine_between are the following:

```xml
<path d="M753.2400512695312,405.12799072265625 l0,-17.5560302734375 l-1.17608642578125,0.156005859375 l0,17.5560302734375 l1.17608642578125,-0.156005859375"/>
```

and 

```xml
<path d="M523.7760009765625,465.91998291015625 l-0.22796630859375,-5.42401123046875 l-0.43206787109375,0.09600830078125 l0.22802734375,5.399993896484375 l0.4320068359375,-0.071990966796875"/>
```

PR #247, opened 2021-10-12
affine_between: return None instead of assert when no significant x part is found

anthrotype:
Fixes https://github.com/googlefonts/picosvg/issues/246

or at least it works around the bug...

```python
from picosvg.svg_types import SVGPath
from picosvg.svg_reuse import normalize, affine_between

s1 = SVGPath(d="M753.2400512695312,405.12799072265625 l0,-17.5560302734375 l-1.17608642578125,0.156005859375 l0,17.5560302734375 l1.17608642578125,-0.156005859375")

s2 = SVGPath(d="M523.7760009765625,465.91998291015625 l-0.22796630859375,-5.42401123046875 l-0.43206787109375,0.09600830078125 l0.22802734375,5.399993896484375 l0.4320068359375,-0.071990966796875")

tolerance = 0.1

assert normalize(s1, tolerance) == normalize(s2, tolerance)

affine_between(s1, s2, tolerance)  # this fails with AssertionError
```

### DIFF
--- a/src/picosvg/svg_reuse.py
+++ b/src/picosvg/svg_reuse.py
@@ -216,7 +216,11 @@ def affine_between(s1: SVGShape, s2: SVGShape, tolerance: float) -> Optional[Aff
     # Align the first edge with a significant x part.
     # Fixes rotation, x-scale, and uniform scaling.
     s2_vec1x_idx, s2_vec1x = _first_significant(_vectors(s2), lambda v: v.x, tolerance)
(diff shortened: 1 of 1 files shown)


anthrotype:
I didn't add a test because in svg_reuse_test.py we currently only test that normalize and affine_between agree with one another (we explicitly assert that if there's no affine between two shapes, then they ought to _not_ normalize the same, which at least in this case it's not true)

PR #250, opened 2021-10-14
normalize: check both first significant vector's components, not only x 

### LINKED ISSUE #249: can't find simple 2x scale affine between two basic triangles
Running this snippet with picosvg==0.18.3 I get this:

```python
>>> from nanoemoji.config import _DEFAULT_CONFIG
>>> from nanoemoji.glyph_reuse import GlyphReuseCache
>>> reuse_cache = GlyphReuseCache(_DEFAULT_CONFIG.reuse_tolerance)
>>> path_a = "M-1,-1 L 0,1 L 1, -1 z"
>>> path_b = "M-2,-2 L 0,2 L 2, -2 z"
>>> reuse_cache.add_glyph("A", path_a)
>>> reuse_cache.try_reuse(path_b)
ReuseResult(glyph_name='A', transform=Affine2D(a=2.0, b=0.0, c=0.0, d=2.0, e=0.0, f=0.0))
```

Instead, with picosvg==0.18.4 as well as 0.18.5 I get `None` result from `try_reuse` function.

It seems to be after #245, still not sure what caused this. 

Related https://github.com/googlefonts/nanoemoji/pull/352

anthrotype:
For normalize() to work as intended, the first vector (with a significant magnitude) should be aligned with Vector(1, 0) -- to eliminate rotation and uniform scaling.
However, the code currently was skipping this transformation when the first vector's .x component was already == 1.0, thus ignoring the .y component. But there are infinite vectors with .x == 1.0 that are _not_ aligned with (1, 0).

Fixes #249

### DIFF
--- a/src/picosvg/svg_reuse.py
+++ b/src/picosvg/svg_reuse.py
@@ -146,7 +146,7 @@ def normalize(shape: SVGShape, tolerance: float) -> SVGPath:
 
     # Normlize first activity to [1 0]; eliminates rotation and uniform scaling
     _, vec_first = _first_significant(_vectors(path), lambda v: v.norm(), tolerance)
-    if vec_first and not almost_equal(vec_first.x, 1.0):
+    if vec_first and not vec_first.almost_equals(Vector(1, 0)):
         assert (
             vec_first.norm() > tolerance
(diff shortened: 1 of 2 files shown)


rsheeter:
Oh ouch good catch

PR #251, opened 2021-10-19
Accept none for clip-path

khaledhosny:
“none” is one of the accepted “clip-path” values:

> https://www.w3.org/TR/SVG11/masking.html#ClipPathProperty

Accept and ignore the clip in this case.

Inkscape sometimes output it (when the object has a clip, then the clip
is released).


anthrotype (approved):
LGTM, thanks!

Issue #257, opened 2021-12-07
explicit stroke="none" is ignored, inherited stroke from parent group overrides it


anthrotype:
when an element has an explicit stroke="none" attribute, this should take priority and stop from inheriting a not-none stroke value of its parent. However, picosvg is ignoring that (probably because "none" happens to also be the default value for the stroke attribute?) and the parent's stroke is inherited despite the explicit stroke="none"

This SVG should display a black square, _without_ any red border stroke:

```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
  <g stroke="red">
    <rect stroke="none" x="10" y="10" width="100" height="100"/>
  </g>
</svg>
```

I noticed this while debugging https://github.com/googlefonts/noto-emoji/pull/373

E.g. take the newly [waved Egypt flag](https://github.com/googlefonts/noto-emoji/pull/373/files?short_path=fe9d148#diff-fe9d148f99f377a2aefe3b6b93265f0b0e05ce9dc4c044ac6f418351296e2e20) built with current picosvg; see how the arabic text inscription looks emboldened, as if a stroke had been applied more than once. In reality, in the original SVG the inscription path has stroke="none" whereas the container has some stroke attribute which inadvertently gets inherited by the children.

Still investigating, however first guesses lead to a couple places, e.g. `to_element` omitting attributes whose value == default, and the `_inherit_attrib` function calling a subspicious `_drop_default_attrib`...

The same issue might occur not just with stroke but with all inheritable attributes actually, whenever a children has an explicit value that's equal to the default which signal that the inheritance chain has to break; whereas we happily ignore defaults and proceed with copying inherited values.

We may have to change the SVGShape classes' default values for all attributes that can be inherited. E.g. stroke currently defaults to "none", however we may have to type it Optional[str] = None to be able to distinguish between the attribute _not being set_ and the attribute being explicitly set to its nomially default value (e.g. "none")...

rsheeter:
See #224 :) Tiresome but a good thing to fix.

PR #258, opened 2021-12-13
Fix explicit attrib defaults being overridden by inherited values

### LINKED ISSUE #224: No setting and explicit default are indistinguishable
After pass from/to element the fill value is dropped so we can't tell that the first one should be black and the second one should inherit fill none. This complicates #223; splitting into it's own issue because I don't think it's worth blocking on.

```
  <g opacity="1" fill="none">
    <ellipse cx="85" cy="50" rx="10" ry="5" fill="black"/>
    <ellipse cx="85" cy="60" rx="10" ry="5"/>
  </g>
```

anthrotype:
Fixes #224 and #257


src/picosvg/svg.py
```diff
@@ -1320,11 +1333,25 @@ def _swap_elements(swaps: Iterable[Tuple[etree.Element, Sequence[etree.Element]]
                 raise ValueError("Lost parent!")
             parent.remove(old_el)
 
+    @lru_cache(maxsize=None)
```
rsheeter, line 1336: Isn't the unbounded cache dangerous?

anthrotype: I don't think so, because I always call cache_clear() at the beginning of _update_etree()
https://github.com/googlefonts/picosvg/pull/258/files#diff-613582736f235e34aeab2643a59d44fb26bfa66e5bcba5f4c6fa195d74b0e7b9R1349

functools.lru_cache(None) is the same as functools.cache(), but the latter is only available for 3.9+
https://docs.python.org/3/library/functools.html#functools.cache

PR #259, opened 2021-12-14
when applying clip path, only use clip-rule for clips and fill-rule for target shape

anthrotype:
clip-rule is only valid for clip definitions within clipPath elements; when we intersect clip paths with the referencing shape, we need to use the shape's fill-rule attribute instead of clip-rule to initialize the pathops.Path on which to perform the intersection.
This fixes an issue with the 'G' (from the motto 'ORDEM E PROGRESSO') of the Brazilian flag in noto-emoji (which is included as a test): here the shape has fill-rule='evenodd' whereas the clipping path has clip-rule='nonzero'; before this fix, we were using the shape's clip-rule (absent, defaulting to 'nonzero'), leading to incorrect rendering as shown below.

Before:
Screenshot 2021-12-14 at 16 18 18

After (correct):
Screenshot 2021-12-14 at 16 18 31


src/picosvg/svg.py
```diff
@@ -573,14 +573,6 @@ def append_to(self, xpath, el):
         self.xpath_one(xpath).append(el)
         return el
 
-    def _combine_clip_paths(self, clip_paths: Sequence[SVGPath]) -> SVGPath:
```
anthrotype, line 576: this was unused, so I deleted it

anthrotype:
for reference, I found about this bug while doing https://github.com/googlefonts/noto-emoji/pull/373

PR #261, opened 2022-01-28
First swing at removal of empty subpaths

### LINKED ISSUE #252: Cleanup empty paths
I have a punch of SVGs that contain empty sub-paths (move followed immediately by close, e.g. `M465,550 Z`) and I’m wondering if picosvg should remove them since they are essentially a no-op? (if one is wondering how I ended up with these, they are anchors and the code I used to generate the SVGs helpfully tried to preserve them by creating such empty paths).

rsheeter:
Fixes #252


src/picosvg/svg.py
```diff
@@ -910,6 +910,23 @@ def round_floats(self, ndigits: int, inplace=False):
...
+            svg = SVG(copy.deepcopy(self.svg_root))
+            svg.remove_empty_subpaths(inplace=True)
+            return svg
+
+        self._update_etree()
+
+        for idx, (el, (shape,)) in enumerate(self._elements()):
+            if isinstance(shape, SVGPath):
+                shape.remove_empty_subpaths(inplace=True)
+                self._set_element(idx, el, (shape,))
```
anthrotype, line 924: since the operation on the shape does `inplace=True`, I think it's sufficient to simply iterate over all `svg.shapes()` and run the `shape.remove_empty_subpaths(inplace=True)`, no need to `_set_elelement`. We do something like that for other inplace operations. Or am I missing something else?

src/picosvg/svg_types.py
```diff
@@ -505,6 +505,32 @@ def callback(subpath_start, curr_xy, cmd, args, prev_xy, prev_cmd, prev_args)
             self._add_cmd(cmd, *args)
         return self
 
+    def subpaths(self) -> Tuple[str, ...]:
+        subpaths = [SVGPath()]
+
+        def subpaths_callback(subpath_start, curr_pos, cmd, args, *_unused):
+            if cmd.upper() == "M":
```
anthrotype, line 512: what if M is preceded by a Z (which I think is quite common)? You don't want to append two empty SVGPaths, one for Z and then another for M. Or maybe you are taking care of removing the empty ones after the callback when you return `tuple(s.d for s in subpaths if s.d)`?

anthrotype (approved):

src/picosvg/svg_types.py
rsheeter: The filter at the end is supposed to dump any completely empty paths

src/picosvg/svg.py
rsheeter: Hm, I think you are right, will amend.

anthrotype (approved):

PR #264, opened 2022-02-01
make subpath moveto absolute before removing empty ones

anthrotype:
When splitting SVGPath into subpaths, we must ensure each subpath's coordinates are independent from the preceding ones. If an empty subpath containing only a moveto is followed by a non-empty one that starts with a relative moveto ('m'), removing the empty subpath without adjusting the relative moveto produces an incorrect result (as shown in the noto-emoji "foot" => https://github.com/googlefonts/picosvg/issues/252#issuecomment-1026839746).

We could call SVGPath.absolute() and then split into subpaths() in order to remove_empty_subpaths() -- inside topicosvg() method, the former occurs after the latter --, but it's preferable that we keep each logically distinct opeartion self-contained and handle relative moveto correctly when we split the subpaths, by making only the first moveto command of each subpath absolute, and keeping the rest of the subpath command unchanged.

### DIFF
--- a/src/picosvg/svg_types.py
+++ b/src/picosvg/svg_types.py
@@ -94,6 +94,12 @@ def _relative_to_absolute(curr_pos, cmd, args):
     )
 
 
+def _relative_to_absolute_moveto(curr_pos, cmd, args):
+    if cmd in ("M", "m"):
+        return _relative_to_absolute(curr_pos, cmd, args)
+    return (cmd, args)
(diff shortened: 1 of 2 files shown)


anthrotype:
the foot is nice and clean now:
Screenshot 2022-02-01 at 16 17 17

rsheeter:
Good catch, ty

PR #267, opened 2022-05-01
Exploratory fix for arc normalization

rsheeter:
Does this make sense to you?

Fixes https://github.com/googlefonts/picosvg/issues/266.

### DIFF
--- a/src/picosvg/svg_reuse.py
+++ b/src/picosvg/svg_reuse.py
@@ -108,6 +108,10 @@ def _affine_friendly(shape: SVGShape) -> SVGPath:
     )
 
 
+ARC_RADIUS_COORD_OFFSET = (
+    -5
+)  # offset from corresponding coord, e.g. x coord is 5, matching radius is 0
+
 # Transform all coords in an affine-friendly path
 def _affine_callback(affine, subpath_start, curr_pos, cmd, args, *_unused):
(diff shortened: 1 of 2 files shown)


src/picosvg/svg_reuse.py
```diff
@@ -129,6 +133,17 @@ def _affine_callback(affine, subpath_start, curr_pos, cmd, args, *_unused):
             new_y = 0
         args[x_coord_idx] = new_x
         args[y_coord_idx] = new_y
+
+        # Arc radii are, excitingly, NOT coords. However, the curvature is entirely different
+        # and nothing normalizes if they are not adjusted so try scaling rx/y proportionally to
+        # the change in magnitude of the respective basis vectors.
+        if cmd.upper() == "A":
+            x_basis = Vector(affine.a, affine.b)
+            y_basis = Vector(affine.c, affine.d)
```
anthrotype, line 142: `picosvg.Affine2D` has `decompose_scale` method that splits an affine into its scale component and whatever remains, we use that in nanoemoji when we need to scale PaintRadialGradient radii. It does basically the same thing.

anthrotype (approved):

src/picosvg/svg_reuse.py
rsheeter: I wonder if I subconsciously remembered that!

Issue #271, opened 2022-05-15
Reuse of arcs is poor


rsheeter:
Shapes like this fail to find reuse:

![0](https://user-images.githubusercontent.com/6466432/168452632-cf816801-9cb0-4594-9dcb-c797d82eb178.svg)

Appears to relate to arcs, specifically arcs whose start/end is almost the same for at least one dimension.

Observed in sanity checking nanoemoji part files for ones that failed (donor is null).

anthrotype:
what's **not** fixed with #272?

rsheeter:
Arcs whose relative endpoint doesn't have a zero coord.

PR #272, opened 2022-06-04
Upgrade arc handling to support circles and ellipses formed of two arcs

rsheeter:
We commonly encounter this in real inputs.

Helps with #271. **Not** as fixing that issue because this only handles the specific subset of the general problem we're actually seeing in practice.


anthrotype:
nice work!

PR #286, opened 2023-02-02
Add the ability to diff

rsheeter:
Handy when hacking on things

### DIFF
--- a/src/picosvg/svg_pathops.py
+++ b/src/picosvg/svg_pathops.py
@@ -129,6 +129,12 @@ def intersection(
     return _do_pathop(pathops.PathOp.INTERSECTION, svg_cmd_seqs, fill_rules)
 
 
+def difference(
+    svg_cmd_seqs: Sequence[SVGCommandSeq], fill_rules: Sequence[str]
+) -> SVGCommandGen:
+    return _do_pathop(pathops.PathOp.DIFFERENCE, svg_cmd_seqs, fill_rules)
+
+
 def remove_overlaps(svg_cmds: SVGCommandSeq, fill_rule: str) -> SVGCommandGen:
     """Create a simplified path filled using the "nonzero" winding rule.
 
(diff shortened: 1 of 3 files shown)


src/picosvg/geometric_types.py
```diff
@@ -44,13 +44,13 @@ def __sub__(self, other: _PointOrVec) -> _PointOrVec:
...
             return self._sub_vec(other)
-        return NotImplemented
+        raise ValueError(f"Other must be a Point or Vector, not a {type(other)}")
 
     def __add__(self, other: "Vector") -> "Point":
         """Return Point translated by other Vector"""
         if isinstance(other, Vector):
             return self.__class__(self.x + other.x, self.y + other.y)
-        return NotImplemented
+        raise ValueError(f"Other must be a Vector, not a {type(other)}")
```
rsheeter, line 53: Changes here are due to typechecker complaining; I don't want the return type to be Point or NotImplemented

anthrotype: but the canonical way to implement math operator overrides is exactly to return NotImplemented when the other type isn't among the ones which the current method is capable of handling, this enables the interpreter to check if the same operator is implemented by the other side of the expression, see https://docs.python.org/3/library/numbers.html#implementing-the-arithmetic-operations

I'm surprised that pytype doesn't automatically support this..

anthrotype:
"Add the ability to diff" -- before reading the code, I thought it was about diffing two svg files..

anthrotype (approved):
LGTM.

by the way, what are you hacking on?

src/picosvg/geometric_types.py
anthrotype: there's even an explicit test for this in pytype test suite:

https://github.com/google/pytype/blob/b6b6448dc562a7200326c92e75efeed203984e16/pytype/tests/test_cmp2.py#L53-L78

rsheeter: Huh. TIL. I will revert this part.

rsheeter:
I wanted to colorize Ewert. Horrible hackery @ https://github.com/rsheeter/colorize. Haven't slapped gradients on it yet but I still quite like the result:

image

PR #287, opened 2023-02-02
Trying to understand what tox is doing for us

rsheeter:
Tox says "if py 3.8 then run pytype and black" (aside: of these only pytype appears to be version-bound). CI says install 3.8 then run tox. Seems like more indirection than we really need.


.github/workflows/ci.yml
```diff
@@ -17,10 +17,12 @@ jobs:
       with:
         # pytype currently doesn't work with python >= 3.9
         python-version: 3.8
-    - name: Install tox
-      run: pip install tox
-    - name: Run style and typing checks
-      run: tox -e lint
+    - name: Install tools
+      run: pip install pytype black
```
anthrotype: the difference is now you are installing these inside the global environment that github CI runner sets up for you, on which you do not have any control. There may already be versions of pytype or black in there, and you are not trying to upgrade them either, nor install any specific versions of them (you should pin the version of these tools as they currently are in 'dev' extra requires in setup.py, otherwise we would get new failures every time we run a CI and upstream tools get updated with new features/bugs). Tox installs using `dev` extra so it will use the pinned versions from setup.py.

rsheeter: Oh we pin there, that makes more sense then.

anthrotype (approved):

.github/workflows/ci.yml
rsheeter: Fixed.

rsheeter:
Closing per further discussion on IM. Key point is that tox is widely known and adopted so experienced Python users will find this helpful.

PR #289, opened 2023-02-22
Made text elements optionally allowed

zond:
- Added `allow_text` options to both `topicosvg()` and `checkpicosvg()`to enable simplifying SVGs with text elements. Note that the text elements will not be converted to paths, they will only pass through.
- Added an `allow_text` boolean flag to the command line tool to control this behaviour.

### DIFF
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -355,6 +355,9 @@ class SVG:
         self.svg_root = svg_root
         self.elements = []
 
+    def _clone(self) -> "SVG":
+        return SVG(svg_root=copy.deepcopy(self.svg_root))
+
     def _elements(self) -> List[Tuple[etree.Element, Tuple[SVGShape, ...]]]:
         if self.elements:
             return self.elements
@@ -403,7 +406,7 @@ class SVG:
     def absolute(self, inplace=False):
         """Converts all basic shapes to their equivalent path."""
         if not inplace:
-            svg = SVG(copy.deepcopy(self.svg_root))
+            svg = self._clone()
             svg.absolute(inplace=True)
             return svg
 
@@ -415,7 +418,7 @@ class SVG:
     def shapes_to_paths(self, inplace=False):
         """Converts all basic shapes to their equivalent path."""
         if not inplace:
-            svg = SVG(copy.deepcopy(self.svg_root))
+            svg = self._clone()
             svg.shapes_to_paths(inplace=True)
             return svg
 
@@ -426,7 +429,7 @@ class SVG:
 
     def expand_shorthand(self, inplace=False):
         if not inplace:
-            svg = SVG(copy.deepcopy(self.svg_root))
+            svg = self._clone()
             svg.expand_shorthand(inplace=True)
             return svg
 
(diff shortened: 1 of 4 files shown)


src/picosvg/svg.py
```diff
@@ -986,6 +991,19 @@ def remove_nonsvg_content(self, inplace=False):
...
 
+    def remove_text(self, inplace=False):
+        if not inplace:
+            svg = self._clone()
+            svg.remove_text(inplace=True)
+            return svg
+
+        self._update_etree()
+
+        for el in self.xpath("//*[name()='text']"):
```
zond, line 1002: I don't understand why SVGs don't allow me to use xpath to find "//text", but according to random bloggers on the web this is the way to do it instead.

anthrotype:
maybe we don't need to add a self.allow_text attribute, how about just add an `allow_text=False` parameter to `topicosvg` method instead?

also, maybe the command line tool could have an additional flag to control that. It's defined [in this module](https://github.com/googlefonts/picosvg/blob/main/src/picosvg/picosvg.py)

anthrotype:
In general, I think we both agree that by default picosvg should not silently drop text elements but raise an error if these are present and the user didn't explicitly request to drop them. And then, it could have an optional mode where one decides to intentionally remove the text elements as part of the picosvg processing. And finally also a way to only remove the text elements (you did that already).
By not storing the `allow_text` in the SVG object state, but instead passing it down as a parameter to `SVG.topicosvg`, I think it makes the patch simpler/smaller (no need to add that `_clone`)

anthrotype:
the title "Added optional support for text elements" may be misunderstood as saying 'picosvg can now simplify/convert text elements to paths' but actually here we are additing support for dropping them, I suggest you reword the PR title accordingly

anthrotype:
actually wait... I got a bit confused.

I think what you want for your use case (from our private conversation) is to retain the text elements as they are, so you just need a way to tell picosvg to not raise an error if they are there. You don't need/want to remove them, and neither we do want to simply remove them. Eventually, one day we may wish to add support for transforming the text to outlines.

So I think we don't need to have a remove_text method at all. We only need a way to allow the text in (with a parameter to `topicosvg` which in turn passes it down to `checkpicosvg`). By default though, we should not allow it, and we can continue to raise an error that text elements are not proper 'pico' svg format.

anthrotype:
technically we could also add support for removing text elements as in deleting them (not in the good sense of converting them to path which we haven't implemented yet), but what purpose would that serve, I'm not sure of it's utility. Unlinke comments or other metadata, the text elements are proper data and picosvg in general doesn't offer ways to "remove this or that". It tries to retain the essence/visual appearance of the original svg while simplifying it's formatting.

zond:
Alright, I believe the best suggestion is to just pass an `allow_text` flag to `topicosvg()` and `checkpicosvg()`, and not have the `allow_text` attribute or `remove_text` method.

I updated the PR. WDYT?

anthrotype:
ok, thanks, looks good that way.
Mind adding a flag to the command line tool as well?

anthrotype:
also please edit the first message (PR description) as it's no longer in sync with the changes, thanks!

anthrotype (approved):

anthrotype:
Thanks

PR #290, opened 2023-02-23
Added support for providing a global bounding box.

zond:
- Added some utility methods to the Rect class.
- Added a bounding_box method to the SVG class.


src/picosvg/geometric_types.py
```diff
@@ -151,6 +151,17 @@ def _overlap(start1, end1, start2, end2):
             return Rect(x1, y1, x2 - x1, y2 - y1)
         return None
 
+    def x2(self) -> float:
```
anthrotype, line 154: how about `x_max` and `y_max`?

we could even make these `@property` getters as addition is cheap; then you will use them like `self.x_max` instead of `self.x_max()` which looks more pythonic

tests/svg_test.py
```diff
@@ -682,3 +682,13 @@ def test_allow_text():
     ):
         text_svg.topicosvg()
     assert "text" in text_svg.topicosvg(allow_text=True).tostring()
+
+
+def test_bounding_box():
+    bounding_svg = load_test_svg("bounding.svg")
+    bounds = bounding_svg.bounding_box()
+    is_close = lambda a, b: abs(a - b) < 1e-5
+    assert is_close(bounds.x, 14.22469)
```
anthrotype, line 691: you could use `math.isclose` function

anthrotype (commented):
thanks, looks good overall, left some minor comments

src/picosvg/geometric_types.py
zond: Done

tests/svg_test.py
zond: Done

zond:
Thanks, good advice!

anthrotype:
thanks!

zond:
:partying_face:

Issue #292, opened 2023-05-07
Pass `tspan` when parsing svg with allow_text=True


NightFurySL2001:
When parsing this svg file to picosvg with allow_text=True, the parser failed at `tspan` element.
`Unable to convert to picosvg: BadElement: /svg[0]/text[0]/tspan[0]`

[會.svg](https://user-images.githubusercontent.com/33471049/236658603-c747a17f-1570-4619-b8e5-f51ce12eacdb.svg)

NightFurySL2001:
Fixed in #294

PR #294, opened 2023-05-08
Allow more SVG text in allowText option

NightFurySL2001:
SVG defineds 3 text element: text, tspan and textPath. Allows all 3 to be passthrough when allowText = True

### DIFF
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -1,6 +1,6 @@
 # Copyright 2020 Google LLC
 #
-# Licensed under the Apache License, Version 2.0 (the "License");/text[0]
+# Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
--- a/src/picosvg/svg.py
+++ b/src/picosvg/svg.py
@@ -1,6 +1,6 @@
 # Copyright 2020 Google LLC
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
+# Licensed under the Apache License, Version 2.0 (the "License");/text[0]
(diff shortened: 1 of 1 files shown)


NightFurySL2001:
Fixes #292

anthrotype:
ok, yes you're right. Add a new test function or parametrize the one for allowText=True, thanks!

anthrotype:
sorry I pushed the wrong button :)

tests/svg_test.py
```diff
@@ -675,14 +675,78 @@ def test_remove_processing_instructions():
...
+                viewBox="0 0 30 30">
+            <text x="20" y="35">
+                <tspan x="0" y="20" style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:10px;font-variant-ligatures:normal;font-variant-caps:small-caps;font-variant-numeric:normal;font-variant-east-asian:normal;stroke-width:1;">Hello</tspan>
+            </text>
+            </svg>
+            """,
+            r"Unable to convert to picosvg: BadElement: /svg\[0\]/text\[0\]",
+            "tspan",
+        ),
+        # tspan without text
```
anthrotype, line 707: is it even valid to have tspan without a parent text element?

NightFurySL2001: Actually, you're right... It's mentioned here: https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Texts

NightFurySL2001:
@anthrotype any updates?

anthrotype (approved):

anthrotype (approved):

anthrotype:
thanks!

PR #295, opened 2023-05-17
Allow vendor style pass through

### LINKED ISSUE #293: Possible to ignore vendor style?
Sometimes SVG generated by Inkscape contains [vendor-specific extensions](https://www.w3.org/TR/CSS2/syndata.html#vendor-keywords) starting with `-`, such as Inkscape's `-inkscape-font-specification` and `-inkscape-stroke`.

NightFurySL2001:
fix #293

### DIFF
--- a/src/picosvg/svg_meta.py
+++ b/src/picosvg/svg_meta.py
@@ -142,7 +142,7 @@ def parse_css_declarations(
     output: MutableMapping[str, Any],
     property_names: Optional[Container[str]] = None,
 ) -> str:
-    """Parse CSS declaration list into {property: value} dict.
+    """Parse CSS declaration list into {property: value} dict. Vendor styles (starting with a dash) are ignored even when provided in property_names as the XML etree module does not support attribute names starting with dash.
 
     Args:
         style: CSS declaration list without the enclosing braces,
@@ -160,6 +160,7 @@ def parse_css_declarations(
     References:
     https://www.w3.org/TR/SVG/styling.html#ElementSpecificStyling
     https://www.w3.org/TR/2013/REC-css-style-attr-20131107/#syntax
+    https://www.w3.org/TR/CSS2/syndata.html#vendor-keywords
     """
     unparsed = []
     for declaration in style.split(";"):
--- a/src/picosvg/svg_meta.py
+++ b/src/picosvg/svg_meta.py
@@ -166,7 +166,9 @@ def parse_css_declarations(
         if declaration.count(":") == 1:
             property_name, value = declaration.split(":")
             property_name = property_name.strip()
-            if property_names is None or property_name in property_names:
+            if (
+                property_names is None or property_name in property_names
+            ) and not property_name.startswith("-"):
                 output[property_name] = value.strip()
             else:
                 unparsed.append(declaration.strip())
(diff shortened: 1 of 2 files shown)


src/picosvg/svg_meta.py
```diff
@@ -166,7 +166,9 @@ def parse_css_declarations(
         if declaration.count(":") == 1:
             property_name, value = declaration.split(":")
             property_name = property_name.strip()
-            if property_names is None or property_name in property_names:
+            if (
+                property_names is None or property_name in property_names
+            ) and not property_name.startswith("-"):
```
anthrotype, line 171: Hm, if `property_name in property_names` then we should include in the parsed output attributes, no matter if it starts or not with `-`. E.g. maybe in the future we might choose to support some of these private attributes, who know.
So instead of `and` you should use `or` for the last condition.

```python
            if property_names is None or property_name in property_names or not property_name.startswith("-"):
```

does that make sense?

anthrotype:
A test would be nice too

anthrotype:
Also maybe please add references to spec e.g. https://www.w3.org/TR/CSS2/syndata.html#vendor-keywords

src/picosvg/svg_meta.py
NightFurySL2001: Just wondering, is there any limitation on why xml etree attribute should not starts with dash? Is it considered a bug or required by the XML standard?

anthrotype: I honestly have no idea, but it's very likely given that lxml is such a popular library, in turn relying on the libxml2 C library. Maybe you could ask on lxml issue tracker.

anthrotype:
> What type of test should I add for this?

in tests/svg_test.py there are some tests for parse_css_declarations function; you could add a couple tests there verifying that the attributes that start with "-" end up in the "unparsed" bit

src/picosvg/svg_meta.py
```diff
@@ -142,7 +142,7 @@ def parse_css_declarations(
     output: MutableMapping[str, Any],
     property_names: Optional[Container[str]] = None,
 ) -> str:
-    """Parse CSS declaration list into {property: value} dict.
+    """Parse CSS declaration list into {property: value} dict. Vendor styles (starting with a dash) are ignored even when provided in property_names as the XML etree module does not support attribute names starting with dash.
```
anthrotype, line 145: the line is a bit too long. Keep the original initial sentence on its own line, the add the rest to the following lines (perhaps keep an empty line separating the first from the rest of the lines) and try to stay within 88 characters if you can

src/picosvg/svg_meta.py
anthrotype: it turns out it's impossible for XML attribute names (or XML names in general) to start with a hyphen, see https://github.com/googlefonts/picosvg/issues/293#issuecomment-1562958015

anthrotype:
thanks, I can pick up from here

NightFurySL2001:
Thanks a lot!

anthrotype:
@NightFurySL2001 https://github.com/googlefonts/picosvg/releases/tag/v0.21.0

anthrotype:
well, yeah.. but it still refers to the issue where we discussed it so it's ok

Issue #297, opened 2023-06-03
Error when removing comment in large SVG file with image


NightFurySL2001:
When parsing this svg file with `topicosvg`, an error occured:
[uni7cb9](https://github.com/googlefonts/picosvg/assets/33471049/d0655eaf-7683-4939-9988-b9cb0eaa3568)
```
Traceback (most recent call last):
  File "d:\Desktop\coding\svg2font\svgbuilder.py", line 209, in build_font_thread
    logs = build_font(
  File "d:\Desktop\coding\svg2font\newfontbuild.py", line 300, in build_font
    glyph, glyphadv = getOutlineFromSVG(svg_fulltext, metrics.upm, metrics.ascender, picoNormalize)
  File "d:\Desktop\coding\svg2font\newfontbuild.py", line 101, in getOutlineFromSVG
    picosvg.topicosvg(inplace=True, allow_text=True)
  File "D:\Desktop\coding\svg2font\kivy_venv\lib\site-packages\picosvg\svg.py", line 1352, in topicosvg
    self.remove_comments(inplace=True)
  File "D:\Desktop\coding\svg2font\kivy_venv\lib\site-packages\picosvg\svg.py", line 1023, in remove_comments
    el.getparent().remove(el)
AttributeError: 'NoneType' object has no attribute 'remove'
```

This seems to be due to the comment on the top of SVG:
`<!-- Generator: Adobe Illustrator 23.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->`

However, interestingly, when the image `xlink:href` and subsequent data is removed in SVG file, `topicosvg` is able to work correctly. 

Using this Python code to remove invalid element like the image that contains the `xlink:href` before parsing with `topicosvg` doesn't help:
```py
    picoerrors = picosvg.checkpicosvg()
    if picoerrors:
        for i in picoerrors:
            # skip if not image and style, prefilter
            if not re.search(r"/style\[\d+\]$", i) and not re.search(r"/image\[\d+\]$", i):
                continue

            # remove style and image elements
            for context in picosvg.breadth_first():
                if re.search(r"/style\[\d+\]$", context.path) or re.search(r"/image\[\d+\]$", context.path):
                    context.element.getparent().remove(context.element)
            break # only need remove once
    picosvg.topicosvg(inplace=True, allow_text=True)
```

rsheeter:
[...]

PR #301, opened 2023-06-07
use XMLParser with remove_comments=True

anthrotype:
Fixes Attribute error when attempting to remove comments with inplace=True and the comment has no parents.
Using the XMLParser(remove_comments=True) option does the trick

Fixes https://github.com/googlefonts/picosvg/issues/297


rsheeter (approved):
lgtm

